PR #93, opened 2015-04-15
Changes the way makeNewValue works to allow fields to replace, not just ...

### LINKED ISSUE #79: Resource form: new/changed  "Omeka Resource" values jump to the bottom
Instead of replacing the input that the user clicked on to change, the Resource field deletes the old input and inserts a new one at the bottom.

This is super-confusing.

jQuery has a simple function to `replace` an existing node with another one, and the resource form code already keeps track of which value is being selected, so this is probably just a matter of adding an argument or otherwise plumbing through the information about which element should get replaced.

kalbers:
...append, on the form. Closes #79.

### DIFF
--- a/application/asset/js/resource-form.js
+++ b/application/asset/js/resource-form.js
@@ -15,7 +15,7 @@
                 },200);
             } else {
                 makeNewField(propertyLi);
-                makeNewValue(qName, true);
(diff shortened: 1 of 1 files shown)


zerocrates:
I'm holding off on this until #94 is done too.

PR #126, opened 2015-04-21
Move ‘add media’ input over to sidebar, adjust new input rows styles for...

kimisgold:
... consistency.

(pinging @jimsafley to look it over)


jimsafley:
The sidebar looks great. My only concern is that media could be added outside the visible page. It's not a huge concern, though.

PR #149, opened 2015-04-29
Adds conditional check for linked resources, and label if there are none...

### LINKED ISSUE #121: On items/show should Linked Resources tab only appear if there are linked resources?

kalbers:
Closes #121


jimsafley:
I forgot, was there consensus on whether we include relationships where the current item is the subject, given that those relationships are already included in the "Values" tab?

kalbers:
I believe we opted for redundancy, but I could be wrong.

Issue #137, opened 2015-04-27
Add display of active filters/search options for browses


zerocrates:
This functionality exists in Omeka 2, and we probably need the same thing in S, if only to just indicate that a search is happening.

PR #155, opened 2015-05-05
Show filters for searching items and item sets. (#137)

kimisgold:


application/src/View/Helper/SearchFilters.php
```diff
@@ -0,0 +1,111 @@
...
+        $partialName = $partialName ?: self::PARTIAL_NAME;
+
+        $escape = $this->getView()->plugin('escapeHtml');
+
+        $filters = array();
+        $exclude = array('submit', 'page', 'sort_by', 'sort_order');
+        $api = $this->getView()->api();
+        $query = $this->request->getQuery()->toArray();
+        $queryTypes = array(
+            'eq' => $escape('has exact value(s)'),
```
zerocrates: All these uses of escape around static strings should be translate calls instead.

The stuff that needs to be escaped is the user input from the request and the stuff coming from the DB (labels of things, etc.), but it's probably easier to just escape that stuff as you're outputting it in the partial.

PR #175, opened 2015-05-19
add oembed whitelist

patrickmj:


application/config/module.config.php
```diff
@@ -247,4 +247,44 @@
             'image/gif'  => 'image',
         ),
     ),
+    'oembed' => array(
+        'whitelist' => array(
+            '#http://(www\.)?youtube\.com/watch.*#i',
```
zerocrates: Seems like these first 6 youtube ones should just be condensed into 3 with the `https?://` construction

application/config/module.config.php
```diff
@@ -247,4 +247,44 @@
...
+            '#http://gi*.photobucket.com/groups/*#',
+            '#https?://(www\.)?scribd\.com/doc/.*#i',
+            '#https?://wordpress.tv/.*#i',
+            '#https?://(.+\.)?polldaddy\.com/.*#i',
+            '#https?://poll\.fm/.*#i',
+            '#https?://(www\.)?funnyordie\.com/videos/.*#i',
+            '#https?://(www\.)?twitter\.com/.+?/status(es)?/.*#i',
+            '#https?://(www\.)?soundcloud\.com/.*#i',
+            '#https?://(www\.)?slideshare\.net/.*#i',
+            '#http://instagr(\.am|am\.com)/p/.*#i',
```
zerocrates: The instagram one doesn't actually seem to work, because while instagram has an oembed endpoint, they don't put the `link` element we use for autodiscovery.

Wordpress handles these cases by also storing the oembed endpoints for things in its list. We could go that way on a case-by-case basis, or just remove the ones that don't work with our current method.

application/config/module.config.php
```diff
@@ -247,4 +247,44 @@
             'image/gif'  => 'image',
         ),
     ),
+    'oembed' => array(
+        'whitelist' => array(
```
zerocrates: All the regexes here probably need start-of-string (`^`) and end-of-string (`$`) anchors to make sure we're matching the whole URL

PR #176, opened 2015-05-19
Allow sidebars to be targeted and fix resource value replacement on resource form. Fix #160.

### LINKED ISSUE #160: Trying to attach resources on item edit opens the delete sidebar on top of the attach item sidebar
The effect looks like the attach item sidebar didn't load, but what's actually happening is that it did load, but both it and the empty delete confirmation sidebar are opening at the same time.

kimisgold:


zerocrates:
Can you separate out the hidden input stuff? As you'd said, that's a separate bug that's on develop, so it doesn't really belong in this pull request.

kimisgold:
Should I go ahead and push the hidden input changes to develop then?

PR #216, opened 2015-06-09
Item set form

patrickmj:


application/view/omeka/admin/item-set/add.phtml
```diff
@@ -8,38 +8,10 @@ $formLabel = $this->plugin('formLabel');
 $formElement = $this->plugin('formElement');
```
zerocrates: Are the various $form helpers still needed on the new versions of the add and edit pages? It doesn't look like they are

application/view/omeka/admin/item-set/edit.phtml
```diff
@@ -4,8 +4,6 @@ $escape = $this->plugin('escapeHtml');
 $escapeAttr = $this->plugin('escapeHtmlAttr');
 $this->headScript()->appendFile($this->assetUrl('js/resource-form.js', 'Omeka'));
```
zerocrates: Similarly, it seems that this line is now unnecessary here, since the script is included by the shared partial.

patrickmj:
Seriously? We're in this project together. One-line fixes should not be delayed this way.

zerocrates:
It doesn't appear to me that this is a one-line fix.

Additionally, line 6 of item-set/edit.phtml appears to be incorrectly retained as well. This, like the lines in my previous comments, jumps out as a red flag because the same line *was* removed from one of the old views but not from the other.

Fixing even small inconsistencies and problems prior to merge, while there's already an explicit channel in the form of this pull request for discussing the changes, seems much preferable to me vs. the options of doing hotfixes myself after the fact, opening new issues for what would be, at that point, one-line fixes, having them be tiny untracked tasks, or simply leaving the problems there.

patrickmj:
What's wrong there in line 6, with the delete button? Looks the same as other views, like item/edit.phtml

zerocrates:
My reference should have been to line 5; line 6 was the line number in the previous version of the file.

patrickmj:
ah. thanks. that makes sense

PR #227, opened 2015-06-11
Config property labels

### LINKED ISSUE #212: Global setting for vocabulary disambiguation
Global admins should be able to configure the information that accompanies property labels on resource show pages:

* show nothing (default)
* show vocabulary (e.g. "Dublin Core")
* show term (e.g. `dcterms:title`)

kalbers:
Allows property label information to be configured from settings page. Closes #212


zerocrates:
I'm not wild about the "magic number" setup of this on the value partial. We just see the values 0, 1, and 2 there without knowing what they are.

I'd replace those with either strings, so they're sort of self-explanatory, or define constants for those values and use those instead of the literals.

PR #237, opened 2015-06-17
Require password

### LINKED ISSUE #235: Users should need to enter their current password before being able to change their password

kalbers:
Closes #235

### DIFF
--- a/application/src/Controller/Admin/UserController.php
+++ b/application/src/Controller/Admin/UserController.php
@@ -127,10 +127,24 @@ class UserController extends AbstractActionController
             $form->setData($this->params()->fromPost());
             if ($form->isValid()) {
                 $values = $form->getData();
-                $user->setPassword($values['password']);
-                $em->flush();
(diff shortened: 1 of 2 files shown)


application/src/Controller/Admin/UserController.php
```diff
@@ -126,18 +131,19 @@ public function changePasswordAction()
...
                 $values = $form->getData();
+                if($currentUser && !$user->verifyPassword($values['current-password'])){
+                        $this->messenger()->addError('The current password entered was invalid.');
+                        return $view;
+                    }
                 $user->setPassword($values['password']);
                 $em->flush();
                 $this->messenger()->addSuccess('Password changed.');
-                return $this->redirect()->toRoute(null, array('action' => 'edit'), array(), true);
+                return $this->redirect()->toRoute(null, array('action' => 'edit'), array(), true);
```
zerocrates: Fix this part here that just adds invalid whitespace and you should be good to go.

PR #246, opened 2015-06-24
Ckconfig

patrickmj:
Any thoughts on
1 how the configuration is pulled in
2. what those toolbars should look like. In the config file, it's pretty easy to move things around as needed/desired

### DIFF
--- a/application/asset/js/ckeditor_config.js
+++ b/application/asset/js/ckeditor_config.js
@@ -1,0 +1,16 @@
+/**
+ * @license Copyright (c) 2003-2015, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or http://ckeditor.com/license
+ */
+CKEDITOR.editorConfig = function( config ) {
+    // Define changes to default configuration here.
+    // For complete reference see:
+    // http://docs.ckeditor.com/#!/api/CKEDITOR.config
+
+    config.toolbar = [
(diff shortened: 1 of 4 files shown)


zerocrates:
What do you mean exactly by the first point? Are there specific alternatives you have in mind?

patrickmj:
I don't have any specific alternatives in mind. That was more a call for whether others know of better approaches.

zerocrates:
I just meant, what does the question mean? "how it's pulled in" meaning the config file vs. direct configuration, or something else?

patrickmj:
Just whether what I'm doing in https://github.com/omeka/omeka-s/blob/ckconfig/application/src/Media/Handler/HtmlHandler.php#L23 and https://github.com/omeka/omeka-s/blob/ckconfig/application/src/Media/Handler/HtmlHandler.php#L69 with the `$view->assetUrl()` business looks right.

zerocrates:
It looks OK to me

patrickmj:
:beers: :+1:

zerocrates:
As for whether the toolbars are right, I seem to remember some discussion with @kimisgold about whether to alter the styles/markup for the Media so the editor could take up the full width. I'd think a yes/no on that point would affect what was acceptable/desired as far as toolbar buttons

kimisgold:
It looks fine to me. The interface is more robust than any of the other ways of adding media, so it looks busier and clunkier comparatively. It might be a matter of finding a better skin, but I can address that later—looks good for merging now.

PR #300, opened 2015-07-13
Site permissions view

kimisgold:


zerocrates:
I don't think I want to merge this without the part(s) of the interface for adding and removing users from that list.

zerocrates:
Are you done with this?

jimsafley:
For other selector controls we use a view helper (e.g. `$this->itemSetSelector()`) that renders a common partial script (e.g. `common/item-set-selector`). I'm not sure it's needed here, but there's already an established pattern and doing so would make the selector reusable.

application/view/omeka/site-admin/index/edit.phtml
```diff
@@ -53,6 +54,83 @@ if ($pages):
...
+                        $usersByInitial[$initial][] = $user;
+                    }
+                ?>
+                <?php foreach ($usersByInitial as $initial => $users): ?>
+                    <li class="selector-parent"><span>
+                        <?php echo $initial; ?></span> 
+                        (<span class="selector-child-count"><?php echo count($usersByInitial[$initial]); ?></span>)
+                        <ul>
+                        <?php foreach($users as $user): ?>
+                            <li class="selector-child" data-child-label="<?php echo $user->name(); ?>" data-user-id="<?php echo $user->id(); ?>" data-user-email="<?php echo $user->email(); ?>"><span><?php echo $user->name(); ?><br><?php echo $user->email(); ?></span></li>
```
jimsafley: Here, in the `data-child-label` attribute, you probably should concatenate `$user->email()` so the user can filter by email as well as user. (Unless there's some reason not to that escapes me.) Also, remember to escape all attributes and user data using `$this->escapeHtml()`;

kimisgold: Ahh okay. I was wondering why the filter wasn't picking up on emails. This probably explains it.

PR #348, opened 2015-08-24
Replaces EscapeHtmlAttr with EscapeHtml.

### LINKED ISSUE #302: Remove usage of escapeHtmlAttr
escapeHtmlAttr doesn't work correctly for all the characters we allow (and have went to some trouble to allow by altering the database schema to allow us to store them).

escapeHtml does not have the same problem with special characters. For any attribute that's contained in quotes (this should be the vast majority if not all attributes already), we can safely use escapeHtml instead with no impact on security.

kalbers:
Closes #302


zerocrates:
Item set form is broken by this PR

PR #463, opened 2015-11-17
Adds alignment options to form and corresponding classes on public vi…

kalbers:
…ew. Also adds thumb size class on public view.


zerocrates:
You probably want to provide a default alignment for the case where one's not in the settings for whatever reason. That's the third getData argument.

zerocrates:
There's a notice about undefined $alignment when you make a new block. You're using that variable always but you only define it when $block exists.

zerocrates:
You may instead want to do something like `$data = $block ? $block->data() : []`, and then use the third argument to getData so you can just move the $alignment assignment out to always happen, rather than having an `if ($block)` as you do now

PR #488, opened 2015-12-11
Checks if form data has changed and warns users if it has.

kalbers:


zerocrates:
All this stuff is only going to work for the first form on a page, I think. I don't know if we need to worry about making it work for pages with multiple forms on them.

PR #504, opened 2016-01-14
Break out sections from site’s ‘general’ form into separate views. (Fix #496)

### LINKED ISSUE #496: Sites: Separate tabs from "General" into pages on left nav
There's a pretty good consensus that it would help to have more direct access to several of those tabs (Navigation, in particular), as well as that they're not really logically connected at the moment.

kimisgold:


zerocrates:
All those new separate forms need CSRF protection. That's why I suggested that while you don't need a new form class for every one of those pages, you probably do want to have an Omeka\Form object for each, because those contain a CSRF element by default and you can just print that in the view and check the form's validity in the action.

zerocrates:
It looks like there may be some styling issue on the Navigation break-out: it looks like there's some spacing issue with the headings in the sidebar.

zerocrates:
A couple issues with the "success" flashes:

* Navigation's just says "site updated" rather than navigation
* The settings page doesn't flash at all

zerocrates:
Actually I guess leave the settings page out of this since you didn't change it in this PR.

Issue #497, opened 2016-01-08
Sites: reflect navigation organization on admin Pages "browse"


zerocrates:
Per feedback: it's difficult to remember or visualize the relationships between pages once you're on the Pages browse, which just presents a flat list in added order.

Sorting the pages in the order they appear in the nav (i.e., in the order they'd be visited by "previous/next" links would be a good start, as well as somehow representing the fact that there's a hierarchy, either with Wordpress-esque hyphenation for children or with a more direct reflection of the Navigation hierarchy.

PR #507, opened 2016-01-20
Model page/browse organization after site navigation to represent ord…

kimisgold:
…er and depth. (#497)


zerocrates:
This doesn't quite work right for order, at least where there are some pages not in the navigation:

nav-hier
page-hier

PR #556, opened 2016-03-16
Add a basic controller test

jajm:
Hi,

this is my attempt to be able to test controllers. I'm not sure if it is the right way, but it's the only way I found.

More description in commit message


zerocrates:
I'm not totally sure if that's the way we want to go for "full" tests either but the changes are small and pretty much okay, so I've pulled them in for now.

Issue #585, opened 2016-04-04
For site blocks, provide a "Select Item" sidebar without media or caption.


jimsafley:
When editing a site page and when adding an attachment to a block, there should be a way to provide the "Select Item" sidebar as usual, but on the following "Attachment" sidebar do *​not* provide the "Media" and "Caption" options. I imagine that there will be certain blocks where the attachment's media and/or caption are irrelevant, so providing them would be confusing to the user.

For example, the "Mapping" module needs to add a "Map" block where a user can select multiple attachments (i.e. items) to display on a map. Each item may have more than one marker. (This is one divergence from the Geolocation plugin, where there's only one marker per item). So, the user assigns media and captions when adding/editing the item, *not* when assigning an attachment to a block.

jimsafley:
Thinking further, we could add a `SiteBlockItem` entity that's similar in use to `SiteBlockAttachment` except that one block cannot contain duplicate items. We could perhaps simulate this restriction within the context of a `SiteBlockAttachment`, but I do have established need for it.

zerocrates:
Having the user mistakenly duplicate items wouldn't actually cause you a problem, would it? I'm not sure it's worth worrying about.

jimsafley:
It's true, I could remove duplicates during the block layout's `onHydrate()`. Even if I don't, it won't have an adverse effect, so you're right, this is probably not worth worrying about.

Still, it would be useful to remove media selection and caption input to avoid confusion.

zerocrates:
The "just pick an item" selector, in some way, seems like it could be useful in a few places, yes.

PR #588, opened 2016-04-05
Provide a way for attachments to be item-only

jimsafley:
Refs #585


zerocrates:
This seems like it would break if you tried to do a "normal" attachment after doing one of these item-only ones.

jimsafley:
I think I've accounted for all circumstances where a user would switch between "item-only" and "normal" attachments. I'm not in love with using `.show()` and `.hide()` to do this, but it's the lowest cost solution, I think.

PR #643, opened 2016-05-26
Add filewriter trait to allow filesystem mocks

stl74:
Allow to mock filesystem easily (needed for  the migration of Archive Repertory module)

### DIFF
--- a/application/src/File/File.php
+++ b/application/src/File/File.php
@@ -92,6 +92,10 @@ class File
         return $this->storageBaseName;
     }
 
+    public function setStorageBaseName($name) {
+        $this->storageBaseName = $name;
+    }
+
     /**
      * Get the Internet media type of the file.
      *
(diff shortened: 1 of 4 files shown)


zerocrates:
It's not really clear to me what this is for if its not being used anywhere.

stl74:
Hi,
I need this for integration tests,.
Example here : https://github.com/biblibre/omeka-s-module-ArchiveRepertory/blob/master/test/Controller/ArchiveRepertoryControllerTest.php (WIP)
If you have another solution to not write on filesystem i'd be pleased to implement it.

zerocrates:
I think it would be simpler to provide an alternate File Store instance entirely that writes to in-memory storage rather than adding a switchable "writer" within the LocalStore. Then whatever features needed for introspection or whatever for testing within the memory-based store can just be confined to that object alone, rather than necessitating modifications to the LocalStore as done here.

stl74:
Ok . I move the localstore mock into biblibre/omeka-s-test-helper

PR #686, opened 2016-07-13
Add setter on Omeka\File

stl74:
I need to change the base name of  the stored file in module ArchiveRepertory.


zerocrates:
Can you add a docblock?

PR #709, opened 2016-08-03
Add page title block

jimsafley:
Don't display the page title in the public view template. Instead, add a block to every new page that displays the page title.

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -21,7 +21,7 @@ class Module extends AbstractModule
     /**
      * This Omeka version.
      */
-    const VERSION = '0.7.1-alpha';
+    const VERSION = '0.7.2-alpha';
 
     /**
      * The vocabulary IRI used to define Omeka application data.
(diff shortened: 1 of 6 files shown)


zerocrates:
@mebrett @kalbers @sheilabrennan 

Thoughts? The idea here is that this adds flexibility: you can remove (or move or duplicate, but not sure that's very likely) the automatic page title heading if you don't want to print the title on some page or want to replace it with something else.

The only downside is that it adds a little boilerplate to all the site page forms: they all start with the "Page Title" block on them as the first one, and it means there's a "Page Title" choice in the right-side sidebar to add to the page.

patrickmj:
If the page title is changed, how would that affect the slug/URL, and hence links to the page?

zerocrates:
This doesn't really have any effect on the title itself, or the slug. It merely lets you move around or remove the visual display of the page title by making it a block. The title is still set and stored where its always been.

sheilabrennan:
Works for me, and think it's a nice option to have for pages.

Issue #699, opened 2016-07-20
Consider using CKEditor's inline editor


zerocrates:
CKEditor allows you to replace your textarea with a contenteditable div rather than the old-school iframe (it can also do contenteditable stuff directly, but textarea replacement would be the easier pathway for us since we have that set up already). This saves some space because the editor toolbar doesn't appear unless the user's actually interacting with that editor. It also means that CKEditor's editor styles don't get used, instead it's just the regular admin page style.

Current:

![oldckeditor](https://cloud.githubusercontent.com/assets/298975/16995874/90320b34-4e7c-11e6-8c72-d72c416ed8a0.png)

Inline:

![inline](https://cloud.githubusercontent.com/assets/298975/16995880/982ce7dc-4e7c-11e6-8f08-7e36c36a0f04.png)

PR #715, opened 2016-08-18
Use and style CKEditor’s inline editor. (#699)

kimisgold:


application/src/Media/Ingester/Html.php
```diff
@@ -97,6 +97,7 @@ protected function getForm(PhpRenderer $view, $id, $value = '')
         $field = $view->formRow($textarea);
         $field .= "
             <script type='text/javascript'>
+                CKEDITOR.inline('$id');
```
zerocrates: Do we need both this line and the following one?

kimisgold: @zerocrates No, I didn't realize I hadn't deleted that second line.

PR #717, opened 2016-08-24
Pull compass out of sass/css. (Fix #710)

### LINKED ISSUE #710: Look into moving off of Compass
Compass seems to be abandoned, or at least abandon-ish.

We should look at alternatives that give us the same benefits but are active. Autoprefixer/PostCSS seems promising as a replacement for the "css3" pieces of Compass, but we'll need to get a strong handle on what other pieces we're using and whether they're provided by something we can drop in, something else we're already using, or a mixin of our own.

This may also mean looking at the Node/Gulp ecosystem instead of Ruby, depending on what we're looking to use. We definitely don't want to add an additional language we simultaneously need people to have installed.

kimisgold:


zerocrates:
Compass is still being used in the advanced-search and jstree files.

zerocrates:
Also, the new property selector sidebar and related styles seem to be messed up

PR #739, opened 2016-09-14
Pass page to block layout form method

jimsafley:
Before, BlockLayoutInterface::form() only took the site object and block object (if available). When the block argument was null there was no way to derive the page object for layouts that need it. In this case, when adding a PageTitle block to a page, it had no way to know the page title.

### DIFF
--- a/application/src/Controller/SiteAdmin/PageController.php
+++ b/application/src/Controller/SiteAdmin/PageController.php
@@ -107,9 +107,14 @@ class PageController extends AbstractActionController
 
     public function blockAction()
     {
+        $site = $this->currentSite();
+        $page = $this->api()->read('site_pages', [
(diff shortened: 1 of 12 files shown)


jimsafley:
If merged any module that adds a block layout will need to update the `BlockLayoutInterface::form()` signature.

zerocrates:
When would the block be null?

jimsafley:
As it was originally, the `$block` arg is null when a block is first added to a page. This change adds the `$page` arg, which is null when the page hasn't been created yet.

zerocrates:
So this would be about what, showing the page title on the admin side? Your specific mention of the page title block, that is. Ah of course I see now it's only touching `form` so it of course is just the admin side.

PR #750, opened 2016-09-20
Make controller plugin Trigger work with Forward

jajm:
Trigger retrieve a RouteMatch object from the Application, but Forward
doesn't modify this RouteMatch. Instead it injects a new RouteMatch
object into the controller. Trigger needs to retrieve this RouteMatch in
order to set the correct identifiers for the SharedEventManager.


zerocrates:
Looks good to me, thanks!

PR #754, opened 2016-09-26
Global settings event

patrickmj:
Moves around the events for adding to global settings page, dividing it into two (one for elements, one for input filters), and making those calls from the forms, not the controller. That means that any modules currently using the one from the controller will need to switch.

Scope moved beyond the advertised global settings page to also include same pattern in site settings.


patrickmj:
Something's still wrong in SiteSettings

patrickmj:
I think all the confusion between modules and filters is sorted out now

zerocrates:
Can we undo all those extra `ViewModel` and whitespace changes in the site controller? Only the Event `use` line change should have been required.

patrickmj:
I suppose. It's just the result of running the fixer on the code. Is that to be avoided?

zerocrates:
I suppose it depends how you're running the fixer. If you're using the rules that are checked into the repository and are used for the tests, then it should be fine. I suspect this is the result of using some other ruleset, probably the defaults.

PR #760, opened 2016-10-03
Allow to filter resources by visibility (is_public)

jajm:


zerocrates:
For consistency with the other filters can you use `$this->createNamedParameter` instead of a literal?

jajm:
> For consistency with the other filters can you use $this->createNamedParameter instead of a literal?

Done. Sorry for the delay.

zerocrates:
Thanks!

PR #762, opened 2016-10-06
Set correct default value of media_type_whitelist and extension_white…

jajm:
…list

Those two settings should be arrays, so this patch changes
FileManager::MEDIA_TYPE_WHITELIST and FileManager::EXTENSION_WHITELIST
accordingly.
It also fixes the installation task where values were inverted

### DIFF
--- a/application/data/migrations/20160930163058_AddFileValidation.php
+++ b/application/data/migrations/20160930163058_AddFileValidation.php
@@ -21,9 +21,9 @@ class AddFileValidation implements ConstructedMigrationInterface
 
     public function up(Connection $conn)
     {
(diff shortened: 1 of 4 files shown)


jimsafley:
Why should the default values be arrays?

jajm:
Both settings are used as arrays (https://github.com/omeka/omeka-s/blob/develop/application/src/File/Manager.php#L405), and the migration script set them as arrays (https://github.com/omeka/omeka-s/blob/develop/application/data/migrations/20160930163058_AddFileValidation.php), so I figured that their default value should be an array too.
But there is a real issue if you do a fresh install. The installation process will set their value as strings and `in_array` will complain.

jimsafley:
Thanks for the changes.

PR #763, opened 2016-10-07
Translate js strings

### LINKED ISSUE #444: Untranslatable strings in globals.js
globals.js directly embeds a few strings that define things like title and aria-label attributes. Since our translation happens entirely in PHP, this makes those strings untranslatable

Examples include the "Search" text in the placeholder of the search bar, the "make public" and "make private" labels, and the "Expand" and "Collapse" labels.

jimsafley:
fixes #444

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -164,7 +164,6 @@ class Module extends AbstractModule
             );
         }
 
-
         $sharedEventManager->attach(
             '*',
             'view.advanced_search',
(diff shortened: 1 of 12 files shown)


zerocrates:
Possibly easier to do the list of strings in merged module config rather than creating a new event?

jajm:
If I understand correctly, you need to write a JS string twice to be able to translate it.
I know it's possible to extract JS strings with xgettext, then convert the resulting PO file into a JS file for use with something like https://slexaxton.github.io/Jed/. Have you considered using something like that ? This would avoid duplicating the strings.

zerocrates:
@jajm You're right: that's an available option for this, and it has some benefits (like not duplicating the strings, as you say).

However, at the very small scale we're dealing with for these at the moment, I think it's simpler and on the whole more beneficial to keep to just the one set of translations all running through PHP. If we get into a situation where the set of strings grows somewhat significantly, we'd have to reconsider and something like what you suggest would probably be a better idea.

Issue #336, opened 2015-08-17
Translate dynamic strings


kalbers:
These lines require manual listing of source strings:

All Vocab, Class, and Proptery labels, specifically vocabs that come with Omeka.

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/item/show.phtml#L24

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/item-set/show.phtml#L34

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/job/browse.phtml#L45
https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/job/browse.phtml#L46

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/job/show-details.phtml#L19
https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/job/show-details.phtml#L31

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/media/show.phtml#L19

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/module/browse.phtml#L129

property/show-details.phtml (multiple)

resource-class/show-details.phtml (multiple)

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/browse.phtml#L32
https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/browse.phtml#L57

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/form.phtml#L19
https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/form.phtml#L22
https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/form.phtml#L24

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/show-details.phtml#L5
https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/show-details.phtml#L9

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/show-property-row.phtml#L10

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/show.phtml#L18

https://github.com/omeka/omeka-s/blob/develop/application/view/omeka/admin/resource-template/show.phtml#L35

[...]

PR #764, opened 2016-10-07
Translate vocabulary, property, and class strings

jimsafley:
refs #336

### DIFF
--- a/application/data/scripts/extract-vocab-strings.php
+++ b/application/data/scripts/extract-vocab-strings.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * Extract labels and comments from bundled vocabularies and output POT message IDs.
+ */
+use Omeka\Installation\Task\InstallDefaultVocabulariesTask;
+
+require dirname(dirname(dirname(__DIR__))) . '/bootstrap.php';
+$config = require OMEKA_PATH . '/application/config/application.config.php';
+$application = Zend\Mvc\Application::init($config);
+$services = $application->getServiceManager();
+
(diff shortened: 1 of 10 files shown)


jimsafley:
One annoying thing `msgcat` does is add a comment containing the name of the temp file created for `extract-vocab-strings`, like so:
```
#. #-#-#-#-#  null1188807331.pot  #-#-#-#-#
#. Property label for Dublin Core:title
#: application/view-admin/omeka/admin/item-set/browse.phtml:6
...
msgid "Title"
msgstr ""
```
This happens only when a message is shared between `extract-vocab-string` and `xgettext` or `extract-tagged-strings`, otherwise it's like so:
```
#. Class label for Dublin Core:Location
msgid "Location"
msgstr ""
```
I see that Omeka classic got around this by combining `Omeka.base.pot` when running `xgettext` using the `--join-existing` flag. Looking at the resulting pot file, it seems the inclusion of the `--add-comments=/` flag prevents the above annoyance from occurring, like so (note the forward slash):
```
#. / Chicago-style item citation: two authors
#: application/models/Item.php:457
#, php-format
msgid "%1$s and %2$s"
msgstr ""
```

zerocrates:
Does `--use-first` to msgcat avoid the extra comments?

jimsafley:
Adding `--use-first` removes the shared `extract-vocab-strings` comments altogether (but not the non-shared ones).

jimsafley:
Note that this is not a bug, just an annoyance.

PR #769, opened 2016-10-14
Add color picker form element

### LINKED ISSUE #765: Create color picker input for theme settings
Default color pickers are different across browsers and OSes, so Omeka needs its own consistent interface for selecting colors for theme settings. At the moment, we're looking to accept hex values and 'transparent', and provide some kind of preview of the color.

jimsafley:
fixes #765


jimsafley:
@zerocrates If it's likely that we'd eventually add alpha channel, shouldn't we store a JSON object that contains the data needed to describe the web color, instead of just a string?

Also, adding color names is trivial. Do you see a reason why I shouldn't add them.

jimsafley:
For the moment I'm going to limit the input and storage to hex and "transparent". If we decide to add alpha channel we could store `rgba(r, g, b, a)`. If it gets any more complex we could also store `{r: 255, g: 255, b: 255, a: 1, foo: bar, baz: bat}`. All we'd need to do is use an element filter to transform user data into the correct storage format, and use a helper to extract whatever format the user needs, like `$this->themeSettingColor('background_color')`.

application/src/Form/View/Helper/FormColorPicker.php
```diff
@@ -0,0 +1,19 @@
...
+
+class FormColorPicker extends FormText
+{
+    public function render(ElementInterface $element)
+    {
+        $view = $this->getView();
+        $view->headScript()->appendFile($view->assetUrl('js/color-picker.js', 'Omeka'));
+        return '
+<div class="color-picker">
+    <div class="color-picker-sample" style="height:2em; border:1px solid #D9D9D9;"></div>
```
kimisgold: Yes.

PR #773, opened 2016-10-18
More events

patrickmj:
I can see lots more places for possible event triggering, but for now this seems to get at the most relevant.


zerocrates:
@patrickmj can you take a look at this again in the wake of the user/edit PR merge?

At a minimum it's now a conflict, but I assume there's some events that need to move around to accomodate the tabbed structure (or perhaps don't need to exist in the same form, since it's mostly built around a Form object).

PR #819, opened 2017-01-09
Fixed database schema for item_item_set.

Daniel-KM:
Hi,

My first commit for Omeka S!

Sincerely,

Daniel Berthereau
Infodoc & Knowledge management

### DIFF
--- a/application/data/install/schema.sql
+++ b/application/data/install/schema.sql
@@ -33,7 +33,7 @@ CREATE TABLE `item_item_set` (
   KEY `IDX_6D0C9625960278D7` (`item_set_id`),
   CONSTRAINT `FK_6D0C9625126F525E` FOREIGN KEY (`item_id`) REFERENCES `item` (`id`) ON DELETE CASCADE,
   CONSTRAINT `FK_6D0C9625960278D7` FOREIGN KEY (`item_set_id`) REFERENCES `item_set` (`id`) ON DELETE CASCADE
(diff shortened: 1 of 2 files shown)


zerocrates:
There's a problem here: that file is automatically generated, so any manual changes such as this one would be overwritten by later updates.

We use Doctrine's tools to generate this file from our annotations on the Entity classes. Doctrine's default is to use `utf8` collations, so we have a little event listener to set the charset to `utf8mb4` for every entity so it generates the way we want. The problem is, that event doesn't affect tables Doctrine automatically creates for many-to-many relationships like this one, so the `utf8` default remains.

Now, in this particular case the error is harmless since there's no text columns in this table anyway. This is likely to hold true for pretty much any many-to-many table in the future as well since we're pretty good about making sure the primary keys for tables are integers. So, I wouldn't be too broken up about not being able to fix this.

To actually fix it in a way that will stick, it'd have to be a change that affected either Doctrine's processing or our scripting around that that generates the schema file. One option that comes to mind that would probably work is to remove the `utf8mb4`-forcing listener entirely and switch to just post-processing the generated schema, finding instances of `DEFAULT CHARSET utf8...` and replacing them with the `utf8mb4` equivalent.

Daniel-KM:
Hi,

I'm just starting to work on Omeka S, so I'm learning Doctrine too now. This is not important for this table, so just a note for later or when a new table requires it really.

Sincerely,

Daniel Berthereau
Infodoc & Knowledge management

PR #825, opened 2017-01-23
Fix/basename file

Daniel-KM:
Hi,

I simplified the way to get the basename in some places.

In fact, the first commit or the second can resolve an issue I have with a module and with names with two extensions (like filename.mets.xml). 

The first commit is basic, but the second one may need some checks, because I don't understand why you didn't choose to use the standard function.

Sincerely,

Daniel Berthereau
Infodoc & Knowledge management

### DIFF
--- a/application/src/File/Manager.php
+++ b/application/src/File/Manager.php
@@ -301,7 +301,7 @@ class Manager
      */
     public function getBasename($name)
     {
-        return strstr($name, '.', true) ?: $name;
+        return pathinfo($name, PATHINFO_FILENAME);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


zerocrates:
For sure the ones that needlessly break apart the stored filename can be changed with no problem, they're just vestiges of an earlier time when we didn't store the storage ID and the extension separately in the database. So, the first commit looks totally fine as you said.

I'll have to see if I can see what the rationale for doing the extension/basename split way is. I _think_ it had something to do with files that had multiple dots in them, but I'm not totally sure about that.

Daniel-KM:
I rebased the first two commits of the patch (the last one was integrated) on the last commit.

Daniel-KM:
De facto merged in beta 5.

Issue #816, opened 2016-12-14
Themes: INI files need omeka_version_constraint


jimsafley:
We need to add `omeka_version_constraint` to every theme's INI file, and respect this constraint when selecting and rendering themes.

jimsafley:
I'm not sure how we should handle a situation like when a *current* theme does not satisfy the version constraint. We could just show an error page on the public side; but what should a site administrator see on the site's theme page? With how we manage themes now, Omeka wouldn't even know the set theme exists. It would be like if someone deleted the theme's directory (which, incidentally, is another problem).

jimsafley:
I think since this introduces the idea of invalid themes (beyond having an invalid INI), I'll look into adding stated registration for themes like we have for modules. I'm still open to easier solutions.

jimsafley:
Fixed in 56db0d4d1c0446df9b8c010e7bddf9ae77f7df7d

PR #832, opened 2017-02-13
Add theme state

jimsafley:
See https://github.com/omeka/omeka-s/issues/816#issuecomment-278925519


application/src/Service/ThemeManagerFactory.php
```diff
@@ -43,11 +48,35 @@ public function __invoke(ContainerInterface $serviceLocator, $requestedName, arr
...
+                $theme->setState(ThemeManager::STATE_INVALID_OMEKA_VERSION);
                 continue;
             }
 
-            $manager->registerTheme(new Theme($dir->getBasename(), $ini, $configSpec));
+            $theme->setState(ThemeManager::STATE_ACTIVE);
+        }
+
+        // Get all themes from the database.
+        $statement = $connection->prepare('SELECT DISTINCT(theme) FROM site');
```
zerocrates: I'd prefer to see us handle the "missing theme" state on a more site-by-site basis.

Meaning, instead of checking all the assigned themes every request to see if any of them are something we don't recognize, do something more akin to our "Fallback" support elsewhere: when we're loading up a site and it requests a nonexistent theme, handle the problem then.

application/view-admin/omeka/site-admin/index/theme-selector.phtml
```diff
@@ -29,5 +35,8 @@
         <input type="hidden" name="o:theme" value="<?php echo $theme->getId(); ?>" <?php echo ($defaultActive) ? '' : 'disabled'; ?>>
     </div>
     <?php endif; ?>
-    <?php endforeach; ?>
+    <?php
+    endif;
+    endforeach;
+    ?>
```
zerocrates: We may want  to display errors here as well...

This goes to part of the weirdness of the theme selector: it's the only place where these themes are viewable, so if we want to show an admin what is wrong so they can fix it or ask the forums, we need to do it here. Otherwise, the failure is silent.

PR #844, opened 2017-02-22
Restyle and mark invalid inputs with '.invalid' class. (#Fix 808)

kimisgold:


zerocrates:
I think I envisioned a slightly different strategy here, but I have to go remind myself of what and why.

zerocrates:
The "touched" logic is specific to `input`... might be a problem with textareas, selects.

PR #848, opened 2017-02-28
Update vocab

jimsafley:


application/src/Controller/Admin/VocabularyController.php
```diff
@@ -115,26 +125,112 @@ public function editAction()
...
             $data = $this->params()->fromPost();
             $form->setData($data);
             if ($form->isValid()) {
                 $response = $this->api($form)->update('vocabularies', $this->params('id'), $data, [], true);
                 if ($response->isSuccess()) {
-                    $this->messenger()->addSuccess('Vocabulary successfully updated'); // @translate
-                    return $this->redirect()->toRoute(null, ['action' => 'browse'], true);
+                    $fileData = $this->params()->fromFiles('file');
+                    if (0 === $fileData['error']) {
+                        $this->messenger()->addSuccess('Vocabulary label and comment successfully updated. Please review these changes before you accept them.'); // @translate
```
zerocrates: I feel like this message is possibly confusing for the usual case where people don't actually update the label and comment. The fact that those changes happen immediately but these go on a delay is just tough to convey clearly.

application/src/Service/RdfImporter.php
```diff
@@ -163,25 +118,156 @@ protected function extractMembers(
...
+
+        // Get classes and properties from the RDF graph.
+        $members = $this->getMembers($strategy, $namespaceUri, $options);
+
+        // Extract differences.
+        $diff = [
+            'classes' => ['new' => [], 'label' => [], 'comment' => []],
+            'properties' => ['new' => [], 'label' => [], 'comment' => []],
+        ];
+        foreach ($members['classes'] as $localName => $info) {
```
zerocrates: I believe these two foreaches are exactly the same, just using different input and output arrays... it's probably worth it to consolidate them to a function that takes those as params.

application/src/Controller/Admin/VocabularyController.php
```diff
@@ -115,26 +125,112 @@ public function editAction()
...
             }
         }
 
-        $view = new ViewModel;
-        $view->setVariable('vocabulary', $vocabulary);
         $view->setVariable('form', $form);
         return $view;
     }
 
+    public function updateAction()
```
zerocrates: We talked about this before you did it, but now seeing it, I don't know how happy I am with the decision to do the updating within the controller.

Minimally I'd prefer the "actually update given this diff" code to live in the Importer and just be called from here.

zerocrates:
Travis failures here are due to the AWS outage

PR #849, opened 2017-03-01
Re-merge view-admin and view-shared directories

### LINKED ISSUE #394: Do we need to allow modules to add admin only views?
To my understanding, only the views-admin directory gets the "only loaded on the admin" treatment. Should we be handling this in config so plugins can have an admin-only views directory too?

zerocrates:
This fixes #394 by undoing our earlier change to separate the "admin" and "shared" views. This change returns us instead to a single view tree.

Since the only conflicting file between the two trees was the layout, the admin layout is moved to a different filename, and a new small MVC listener switches the layout to that one when we're in the admin. Otherwise all views are the same and all available on both "sides."

This is a fairly large change just in terms of moving files around, so requesting the attention of @jimsafley @patrickmj @kimisgold


patrickmj:
The sidebar for adding an attachment for an Item with Metadata appears confused
additemwithmetadata

zerocrates:
Hmm that's an interesting one, I'll take a look...

zerocrates:
For some reason `setTerminal` which says to not use the layout doesn't work if you use it after you make a call to change the layout. So that's fun.

zerocrates:
@patrickmj should be fixed

patrickmj:
yep, looks fixed to me, too

patrickmj:
Omeka2Import, CSVImport, and Sharing (the one I was most worried about) seem fine so far with the changes here.

PR #854, opened 2017-03-08
Import template

jimsafley:


application/src/Controller/Admin/ResourceTemplateController.php
```diff
@@ -40,6 +50,271 @@ public function showDetailsAction()
...
+
+        if ($this->getRequest()->isPost()) {
+            $data = $this->params()->fromPost();
+            $form->setData($data);
+            if ($form->isValid()) {
+                $file = $this->params()->fromFiles('file');
+                if ($file) {
+                    // Process import form.
+                    $import = json_decode(file_get_contents($file['tmp_name']), true);
+                    if (JSON_ERROR_NONE === json_last_error()) {
```
zerocrates: Probably a little unnecessary complexity here... json_decode will return null on error. Sometimes you need to check for errors explictly but since you already only allow arrays as "valid" data it seems you could collapse this into just checking `importIsValid` and relying on the null being rejected by the check against array there.

This would let you drop an indentation level and a duplicated `else` clause.

application/src/Controller/Admin/ResourceTemplateController.php
```diff
@@ -40,6 +50,271 @@ public function showDetailsAction()
...
+            // missing or invalid label
+            return false;
+        }
+
+        // Validate class.
+        if (!isset($import['o:resource_class']) || !is_array($import['o:resource_class'])) {
+            // missing or invalid o:resource_class
+            return false;
+        }
+        if (!array_key_exists('vocabulary_namespace_uri', $import['o:resource_class'])
```
zerocrates: Does this prevent import of templates that don't have an assigned resource class?

PR #859, opened 2017-03-16
Fix/render iiif

Daniel-KM:


zerocrates:
I seem to remember there being a reason for having a defined width for the openseadragon viewer... that it didn't play nicely with responsive resizing? Maybe I'm misremembering or thinking of an older version.

Daniel-KM:
The height is required, else the div will be zero sized. The width is not required, by default a div takes whole available size.

zerocrates:
The unconstrained width does indeed seem to work fine and respond to resizes, so we're okay there.

On the noscript message, the `// @translate` comment only is intended for usages where it's passed into some remote call to the translator (so the `xgettext` command can't tell that the string is translatable). Here, it's more appropriate to directly use the message string in a call to `$view->translate`. As it is, it won't actually be translated.

Otherwise this looks good to me!

zerocrates:
Thanks.

Issue #809, opened 2016-11-28
More flexibility about what the API manager does


zerocrates:
A decent amount of the trouble encountered doing more "interesting" or intensive with the API is that it's pretty rigid in what it does and what it allows you to change. Ultimately for most things the core of what it does is find entities and hydrate them, but it generally mandates that it returns Representations, not Entities (and not References either), and that it flushes at every step.

Mostly this makes sense for the normal flow of things: we want to use the Representations generally, and we need the flush to occur so new data like the IDs of newly created records will exist so the Representation can be created. In jobs and bulk processing and a few other places, this doesn't work though: we need the Entity so we can detach it, or so we can do things that aren't in the API, or we want to reduce the number of expensive flush() calls.

I think it makes sense to try to let the API accomodate some of these other use cases. The API could

- return stuff other than Representations (like References instead, or even Entities)
- detach the entities that it returns
- just hydrate, not flush, leaving the API consumer the duty of flushing (or not)

None of these are desirable in the normal case, but we have current needs for all of these things and currently we're either doing hacks to make it happen or we're not using the API at all (or we build in a backdoor to get Entity access from the Representation, as in the case of Users).

jimsafley:
Some possible issues with the changes currently in #861:

First, we are no longer validating API response content. I wonder if we should by checking against the three "responseContent" request options. (Or something more flexible but complicated.)

R elatedly, listeners to `api.*.post` and `api.execute.post `can no longer rely that the passed response contains objects that implement `RepresentationInterface`. They must check the "responseContent" request option to know. To guarantee a return type, it may be in our best interests to transform `Api\ResourceInterface` content to `Api\Representation\RepresentationInterface` within `finialize()`.

jimsafley:
To resolve the above concerns, adapters must now set `Api\ResourceInterface` objects to the resource content. The manager will transform the API resources (respecting the the "responseContent" request option) in `finalize()` _after_ triggering the `post` events.

jimsafley:
Above concerns resolved here 8d0ba562cf4d5abbcb2bcbce81e28d2a7c9160e7

jimsafley:
Resolved in #856 and #861

PR #861, opened 2017-03-19
Refactored API

jimsafley:
See #809


jimsafley:
This PR will break backwards compatibility. Obviously I need to fix tests before this can be merged.

### Notable changes

- API requests and responses no longer extend `Zend\Stdlib\Request|Response`
- The API manager no longer catches `ValidationException` (The API consumer must now handle them)
- The API manager returns only successful responses
- The API response no longer stores state or `ErrorStore`
- On `create`, `batch_create`, and `update` the API controller plugin returns `false` by default during a `ValidationException`
- Added 2nd argument to the API controller plugin's `__invoke()`: `$throwValidationException` (throws a  `ValidationException` instead of `false` on validation error) 
- Added `prepareResponseContent()` to the API's abstract adapter (adapters should pass their API resources through this before setting to the response)
- API request options must be set using `$request->setOption()`
- API request options:
  - initialize: (bool) Set whether to initialize the request during execute() (e.g. trigger API-pre events). Default is true.
  - finalize: (bool) Set whether to finalize the request during execute() (e.g. trigger API-post events). Default is true.
  - isPartial: (bool) Set whether this is a partial UPDATE request (aka PATCH). Default is false.
  - continueOnError: (bool) Set whether a BATCH_CREATE operation should continue processing on error. Default is false.
  - flushEntityManager: (bool) Set whether to flush the entity manager during CREATE, UPDATE, and DELETE. Default is true.
  - responseContent: (string) Set the type of content the API response should contain. Default is "representation". The types are:
    - representation: an API resource representation (implements `Omeka\Api\Representation\RepresentationInterface`)
    - reference:  an API resource reference (instance of `Omeka\Api\Representation\ResourceReference`)
    - resource: an API resource (implements `Omeka\Api\ResourceInterface`)

### Backwards incompatible breaks:

- Listeners to `api.*.post` and `api.execute.post` can no longer rely that the passed `Api\Response` contains objects that implement `RepresentationInterface` (must check `Api\Request::getOption('responseContent')` to know)
- Removed `Api\Manager::executeBatchCreate()` (adapters implementing `batch_create` must now call initialize and finalize events themselves.)
- Removed `Api\Request::continueOnError()`
- Removed `Api\Request::isPartial()`
- Removed `$continueOnError` as 4th argument in `ApiManager::batchCreate()`
- Removed `$partial` as 5th argument in `Api\Manager::update()`
- Removed `$partial` as 5th argument in `Controller\Plugin\Api::update()`
- Removed `Api\Response::setStatus()`
- Removed `Api\Response::getStatus()`
- Removed `Api\Response::mergeErrors()`
- Removed `Api\Response::addError()`
- Removed `Api\Response::getErrorStore()`
- Removed `Api\Response::getErrors()`
- Removed `Api\Response::isError()`
- Removed `Api\Response::isSuccess()`

PR #871, opened 2017-04-05
Fixed check of slugs.

Daniel-KM:
The check of slugs is not the same in adapters and in routes.config.ini lines https://github.com/omeka/omeka-s/blob/develop/application/config/routes.config.php#L21 and next ones.


zerocrates:
Can you add "underscores, " to the error texts also, so it accurately states the allowed set?

zerocrates:
Thanks!

Daniel-KM:
Done.

zerocrates:
For some Github thinks your changes are from 19 hours ago so they're sorting oddly on here... maybe a clock or timezone problem on your system? Not a problem, just interesting.

PR #872, opened 2017-04-06
updated composer to include the composer merge plugin, This will allow modules to have there own dependencies

elrossco22:
updated composer to include the composer merge plugin, This will allow modules to have there own dependencies

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,8 @@
         "zendframework/zendxml": "^1.0.2",
         "omeka/composer-addon-installer": "*",
         "omeka-s-themes/default": "dev-develop",
-        "beberlei/DoctrineExtensions": "^1.0"
+        "beberlei/DoctrineExtensions": "^1.0",
+        "wikimedia/composer-merge-plugin": "dev-master"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.2",
@@ -74,6 +75,17 @@
     },
     "autoload": {
         "psr-4": {"Omeka\\": "application/src"}
+    },
+    "extra": {
+        "merge-plugin": {
+            "include": [
+                "modules/*/composer.json"
+            ],
+            "recurse": true,
+            "replace": false,
+            "merge-dev": true,
+            "merge-extra": false
+        }
     },
     "repositories": [
         {
--- a/composer.lock
+++ b/composer.lock
@@ -4,20 +4,21 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "b7475b0a64dade1b2237b6da0eac393a",
+    "hash": "040e074094f001e45f08e1c460acbf6e",
+    "content-hash": "dc84c02c2fea9540d56068c5c7e11ecc",
     "packages": [
         {
             "name": "beberlei/DoctrineExtensions",
-            "version": "v1.0.13",
+            "version": "v1.0.15",
             "source": {
                 "type": "git",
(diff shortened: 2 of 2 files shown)


elrossco22:
@zerocrates We've been developing a range of modules for the OS community and have found frequent issues when creating modules which have there own dependencies, The above approach is popular with the likes of Wikimedia, Drupal etc

zerocrates:
This basically makes users installing a module install it with Composer, right? (at least if it has Composer dependencies) We've tried to avoid that being the case: though we're pulling in a bunch of dependencies with Composer, the packages we publish have the dependencies preinstalled so users don't need to use the CLI.

Can you elaborate on the kind of issues you're encountering? It would help to have some specifics so we know what the costs of the current setup are. We might be able to suggest or come up with some alternate solution as well. We're already doing a small Composer plugin ourselves to handle pre-bundled themes/modules.

elrossco22:
Hi, Not quite. 

Users install modules as per normal (no need to install them with composer), The user then runs `composer update` in the project root (not the module dir) the composer merge plugin in the PR is set up to traverse with the following pattrern `"modules/*/composer.json"` so will find all of said modules with there own composer.json file and download the dependencies to the Omeka vendor folder in root, it will also regenerate the projects autoload to make the modules classe's and namespaces available.

The issue we have at present is that we are developing an OAuth Authentication module for Omeka-S, rather than re-invent the wheel we are leaning of work already done by the phpleague (https://github.com/thephpleague/oauth2-client) and as such this is more of an Omeka-s wrapper. As part of our module we must package up the work of the phpleague and find a suitable way to install its dependancies

zerocrates:
Thanks, that's a helpful clarification. In some ways it'd actually be better if it _did_ require the module to be installed with Composer, as it _does_ require the end-user to re-run Composer to download the deps and regen the autoloader.

My instinct/preference for modules with their own dependencies to ship them themselves: have their own vendor folder, their own autoloader, etc. It's definitely swimming upstream to try to get Composer to behave this way though.

Does just using the module's composer.json directly work (meaning dependencies are installed under a vendor folder within the module)? Or have you encountered problems trying that? I believe the main issue with the naive approach is that you'll get duplicate installation of things already provided by the S core, though this doesn't cause an error apart from that as far as I'm aware. When I'd thought over this previously, my plan was to try to do something in the realm of Composer's "provides" or something similar to get it to skip any dependencies that were already present in the core.

Moving to a workflow that requires users to interact with Composer is possible and definitely makes some of this simpler, but I'd like to avoid it if we can reasonably do so.

elrossco22:
I think the best course of action for this would be for me to prepare Gist's or PR's with each of the available options and we can discuss the pro's/con's of each. 

I also think this is part of wider discussion in regards to Omeka-S's direction, as the options discussed above will ultimately lead to Omeka being a more developer focussed platfrom in regards to modules being installed unless there is some sort of "plugin installer" which would automate running composer etc for a end non technical user.

zerocrates:
Well, my goal with my option I suggested was to avoid making users use
Composer: the module can ship with the vendor folder already filled, just
like the core does.

I agree though that the other options, and really anything approaching
Composer's standard workflow, would tilt the balance of S's requirements on
a user installing or administering it.

Running things though Composer through the interface as you suggest is an
option, but it requires things like having the vendor and add-ons folders
web writable, a road I'd rather not go down.

Anyway, I agree that concrete comparison of different options is probably
best here.

zerocrates:
I don't think this ever ended up moving anywhere closer to a resolution, so it's probably best to close it.

Dependencies for modules is still a somewhat tricky situation in places, but I don't know that there really are options that don't involve just using Composer at the "top" level so to speak, which we're still leery about mandating for users.

PR #879, opened 2017-04-17
Slugify non-ascii characters.

Daniel-KM:
Slugify "café" to "cafe", not "caf-".

### DIFF
--- a/application/src/Api/Adapter/SiteSlugTrait.php
+++ b/application/src/Api/Adapter/SiteSlugTrait.php
@@ -62,8 +62,12 @@ trait SiteSlugTrait
      */
     protected function slugify($input)
     {
-        $transliterator = \Transliterator::createFromRules(':: NFD; :: [:Nonspacing Mark:] Remove; :: NFC;');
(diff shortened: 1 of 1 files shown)


zerocrates:
The only problem here is that this ends up mandating the intl extension, something we've tried to avoid as it's missing on lots of hosts (yes, even though it's bundled with PHP since forever).

We use it ourselves other places, but it tends to be in ways that we can just skip it if it's not there (as in the date processing code).

Doing this with a lookup table or maybe iconv with TRANSLIT are possible alternatives, either altogether or in the case that the Transliterator class can't be found. Simply skipping the transformation in that case could be fine too.

Daniel-KM:
Just added a check for the extension and added an alternative with iconv. iconv doesn't convert "é" and some common characters, but it converts some others ones.

zerocrates:
Just for extra safety, we should probably stick the `iconv` behind a check that the `iconv` extension is loaded as well.

zerocrates:
Thanks.

PR #890, opened 2017-04-21
Added a trigger to view details in the resource show pages.

Daniel-KM:
The name of the trigger may be changed to avoid the same event than in browse view.


zerocrates:
@patrickmj what do you think?

patrickmj:
I'm generally a fan of adding more versatility to our interfaces, so I have no concerns. Just some convo about naming conventions and possible evolution is in order.

Daniel-KM:
I'm ending a folksonomy module and it's better to see the one or two tags of a resource in the sidebar than in a new tab. Nevertheless, to avoid to mix with the browse show-details (the events are attached to all the controller, not only one action), the name may be "view.sidebar".

zerocrates:
Since it's show-page specific, I think we'd prefer `view.show.sidebar`.

Daniel-KM:
The module [Folksonomy](https://github.com/Daniel-KM/Omeka-S-module-Folksonomy) is nearly ended, but don't use the sidebar currently. The module [CleanUrl](https://github.com/Daniel-KM/Omeka-S-module-CleanUrl) will use it on the next update.

PR #892, opened 2017-04-25
Fixed resource reference from api.

Daniel-KM:


zerocrates:
Is there a bug this fixes?

Daniel-KM:
Yes, this is a bug fix. It is needed for the module [Folksonomy](https://github.com/Daniel-KM/Omeka-S-module-Folksonomy). The id of the tag is its name, so I need to change the resource reference, so the reference should be determined from the resource, not from a new object.

Issue #868, opened 2017-03-31
MissingDependencyModuleException


stephenwf:
Modules (`Module.php`) in Omeka that implement the interface `Zend\ModuleManager\Feature\DependencyIndicatorInterface` can specify Zend module dependencies.

Example module:
**Module.php**
```php
<?php

namespace MyModule;

use Omeka\Module\AbstractModule;

class Module extends AbstractModule implements DependencyIndicatorInterface
{
    public function getConfig()
    {
         return [];
    }

    public function getModuleDependencies()
    {
        return ['ZfcTwig'];
    }
}
```

**Expected when on module page:**
Some message explaining that the module cannot be installed, or graceful message when installation fails.

**Actual**
Tries to install and throws: `Zend\ModuleManager\Exception\MissingDependencyModuleException`

zerocrates:
Hmmm... we haven't used this ourselves, which is why Omeka doesn't look for that exception. I'd have to do some looking at what our options are here.

PR #904, opened 2017-05-04
Added module weighting system

elrossco22:


elrossco22:
@zerocrates @patrickmj @stephenwf @apricot13 @euanfergusson As previously discussed on our call this should allow modules to have a simple weighting system

zerocrates:
Can you refresh my memory as to what your specific problem related to module load order was? As in, what's the error or problem that's occurring due to different ordering? I seem to remember it was something about configuration merge precedence, right?

It occurs to me that it might be a better solution to that specific problem to use Zend Framework's existing events for manipulating the merged config (see https://framework.zend.com/manual/2.4/en/tutorials/config.advanced.html#manipulating-merged-configuration ). i.e., the configuration that should take precedence can be specified by a EVENT_MERGE_CONFIG listener instead of "regular" module configuration, and therefore always "win," regardless of the module load order.

My reluctance to this specific implementation is that it's in the database, and therefore relies at the moment on manual manipulation of the database to have any effect. A UI could be added but then there's a UI for a fairly niche feature. If specified load order is still necessary, it might be preferable to specify it in a file as that's incrementally more accessible.

elrossco22:
Hi @zerocrates This fixes a slightly different issue to the one you've described above. We have 2 modules module A and module B, Module B builds on top of existing functionality provided by Module A and has Module A  as a dependancy as such. For module B to load and function we have to ensure that module A is loaded prior to it by the module manager

zerocrates:
So the problem is just that the dependency check fails (i.e., #868)?

elrossco22:
@zerocrates we are able to achieve dependancies with code as follows `
public function getModuleDependencies()
    {
        // You can use this to enforce Twig in an installation if its required.
        // See open bug (https://github.com/omeka/omeka-s/issues/868)
        // return ['ZfcTwig'];
        return ['IIIFIntegration'];
    }`

However as we're going to be Open sourcing multiple modules which can be combined together and cant guarantee which subset of the modules a user will install and use we cant hard code these values

zerocrates:
Okay so, by specifics, what I'm interested in is exactly what error or problem happens and where it occurs, in the absence of this feature.

You've told me the relevant problem isn't about configuration merge order, nor is it about declared module dependencies failing. So, what _is_ the problem (or at least one concrete example).

The reason I'm wondering is that the usual order of things by the Zend module system is such that the order isn't significant (since most/all work happens in events fired after all modules are loaded, and classes/etc. are autoloaded), so I'm interested to know where that's not the case.

My preference is for there to be a solution to the underlying problem that doesn't involve weighting like this. In the event that weighting is in fact necessary, it's probably going to end up as more useful to specify those weights somewhere other than the database (like each module's ini file, for example).

stephenwf:
@zerocrates 

One example is a module that has `templateA` and `templateB` in the module. Now these contain "defaults" or reference templates for custom modules to override. The intention is that a theme or a module can simply chuck their template into folder `view/myModule/templateA` and enable the module. At this point Zend should load `templateB` from the first module, and `templateA` from the second module. Now apply the same problem to ServiceManager/DI configuration overrides and you can begin to see our problem.

Now this is the case depending on which order the modules are the database, but this is not guaranteed. What we are looking for is deterministic loading order. I think another solution would be to load modules depending on their dependants. Something like how this library works: https://packagist.org/packages/bulton-fr/dependency-tree

If the database stored the dependencies, it could build a tree and a reliable load order, no changing the database required

zerocrates:
So:

The explicit example you're offering really is a configuration merge order issue (as it's the order that the modules add themselves to the template path stack that's relevant for which gets picked), so that seems like it actually is solvable with an EVENT_MERGE_CONFIG-type solution rather than tinkering with the load order. As for ServiceManager configuration issues, that's probably the same kind of situation, and the ServiceManager also offers some methods of its own for modifying services, but if you truly want the dependency to be optional those could be tough to use. It would depend on the specifics.

Back to the load order stuff itself though: Declaring dependencies (and/or "soft" dependencies, since it seems that that's what you're looking for here) and deciding the load order based on those is a better model in my mind, because it doesn't require coordination among different modules or developers on the weights they choose, or manual tinkering with the order by end-users. Once again though, as with the simple weight option you're presenting here, I believe those should be stored in the modules' .ini files, not the database. We'd need to source them from somewhere to get them into the database in the first place anyway, and the order logic is unlikely to be happening in the database itself (as it doesn't in this PR... it could be easily rewritten to do so, but a dependency chain or tree less so), so we might as well not mess with the database storage for little if any benefit.

Daniel-KM:
See my comment on #1091.

Issue #907, opened 2017-05-05
Confusing CSRF error message when upload item is too large


crism:
If a user tries to upload media that exceeds some limit (I believe the limit set in `php.ini`), the upload page displays “CSRF: Value is required and can't be empty,” which only serves to confuse the user.

zerocrates:
The reason this happens is that when you go over the post_max_size, PHP just drops all the data out of the request entirely. We see this as you missing the CSRF token on the page and so we throw that error.

PHP doesn't really tell us in a useful way that you've gone over the limit, but we may be able to detect this in some other way.

crism:
In `ItemController::addAction()`, the form data is populated with `$this->params()->fromPost()`. However, since PHP has already rejected the post at a lower level, that comes back empty. Validation then fails on the first required field.

(There is a warning to the user on the Media upload form, but it’s hidden by default.)

crism:
Ah, @zerocrates—crossed signals. `(-:` My first preference would be to see if we can detect that PHP has rejected the post as too large—but if not, give a more meaningful message when a post comes in apparently with no data, with a diagnostic suggestion.

crism:
This fix inexplicably[*] causes user account creation to fail:
```
[Thu May 18 11:46:15.597082 2017] [:error] PHP Notice:  Undefined index: user-information in .../omeka-s-dev/application/src/Controller/Admin/UserController.php on line 45, referer: .../omeka-s/admin/user/add
[Thu May 18 11:46:15.597135 2017] [:error] PHP Catchable fatal error:  Argument 2 passed to Omeka\\Api\\Manager::create() must be of the type array, null given, called in .../omeka-s-dev/application/src/Mvc/Controller/Plugin/Api.php on line 94 and defined in .../omeka-s-dev/application/src/Api/Manager.php on line 71, referer: .../omeka-s/admin/user/add
```
I’m looking into it…

[*] Nothing is inexplicable, but I sure don’t understand this yet.

zerocrates:
My preference for fixing this would be to move to a higher layer to detect the issue, somewhere like a MVC controller/dispatch event listener, where we can just check if the current request is a POST with no data submitted, then do something like throw an exception if so (optionally checking to see if there's a request length, as I believe that remains intact even after PHP trashes the post data).

[...]

PR #908, opened 2017-05-05
omeka/omeka-s#907: adding error message if no post data received

crism:
Not wedded to this wording, but it does the job.

### DIFF
--- a/application/src/Controller/Admin/ItemController.php
+++ b/application/src/Controller/Admin/ItemController.php
@@ -206,7 +206,9 @@ class ItemController extends AbstractActionController
         if ($this->getRequest()->isPost()) {
             $data = $this->params()->fromPost();
             $form->setData($data);
-            if ($form->isValid()) {
+            if (count($data) <= 0) {
+                $this->messenger()->addError('No form data received. Perhaps a file was too large?'); // @translate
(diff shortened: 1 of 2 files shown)


zerocrates:
I think I'd like to try to solve this in a way that it's solved across the whole UI, rather than having to deal with it piecemeal anywhere where there's a POST. (Even though file uploads are the most common cause, they're not the only one).

crism:
I’m not sure there’s another place to put it; this pattern could easily be applied elsewhere. It’s only relevant where there’s already a check for `isPost()`, and seems a lot easier than asking the `Form` to try and figure out if its fields are all empty after they’ve been set. Although, hm, the check could be in `setData()`.

crism:
Ah… a bigger problem is that any really global changes would be in `Form` itself, which is part of ZendFramework, and so a much more radical change than anything restricted to Omeka S. The change could go in `ResourceForm` in this case, but it would need to be similarly implemented in each appropriate subclass of `Form`…

crism:
@zerocrates: This seems to handle the problem. As before, not wedded to the exact wording, but this seems a heck of a lot clearer to confused users.

crism:
As per #907 discussion, this is probably not the right approach and has unpleasant side-effects.

PR #911, opened 2017-05-08
Added a trigger to filter search filters.

Daniel-KM:
For [Folksonomy](https://github.com/Daniel-KM/Omeka-S-module-Folksonomy) too, even if it can work without it.


zerocrates:
I see no problem with this, but we might be changing the structure of the search filters array here in the near future to move the filter label out of the key (as it's currently problematic for the way the advanced search now works, plus it means the keys are affected by the active language).

Just something to be aware of.

PR #916, opened 2017-05-11
catch_connection_exceptions

elrossco22:
In the event MySQL goes away for any period of time, the full stack trace is outputted to all users including the MySQL credentials .. password etc.See screenshot below the password "Password123" is exposed.
[https://postimg.org/image/3zk2ymooh/](https://postimg.org/image/3zk2ymooh/)

### DIFF
--- a/application/src/Service/ModuleManagerFactory.php
+++ b/application/src/Service/ModuleManagerFactory.php
@@ -100,6 +100,11 @@ class ModuleManagerFactory implements FactoryInterface
             // If the module table is not found we can assume that the
             // application is not installed.
             $status->setIsInstalled(false);
+            
+        } catch (ConnectionException $e) {
+            // If the module table is not found we can assume that the
+            // application is not installed.
+            $status->setIsInstalled(false);
         }
 
         foreach ($dbModules as $moduleRow) {


zerocrates:
Well that's definitely not for the best...

That fallback exception screen is pretty much a temporary placeholder. The standard "classic" answer to this problem is to not show any error details at all for any exception unless "development" mode is activated in the .htaccess.

Special handling for database exceptions is fine as well, though what does this PR result in for the user? It would seem like it'd cause a redirect to the installer, which would be a pretty confusing error experience.

elrossco22:
Screenshot below of the result of this PR, All being perhaps a little confusing as it does say "Install"

https://postimg.org/image/td6xkhdj1/

We could create a custom page "site under maintenance" etc and have that displayed instead

elrossco22:
@zerocrates Are there any thoughts in regards to my last comment, I'd be keen to wrap this up as its a security flaw ?

zerocrates:
So to start what I've done is just restored the mode of operation from Classic: exception messages don't get printed to users at all unless the administrator has affirmatively set "development" mode on. So in the case of connectivity issues by default all that will now be seen is this:

error

elrossco22:
@zerocrates Great, Thanks for the update. Whats the plan for this ? Is there a case for not using the Classic exception messages ?

zerocrates:
Clarify for me: is the change I just mentioned of blanking the details by default, which I committed just before commenting, sufficient, or are you looking for something further?

elrossco22:
@zerocrates Sorry just seen this, The blanking of details is fine for now.I'll close this

PR #922, opened 2017-05-15
Bugfix/duplicate initial

euanfergusson:


zerocrates:
My comments:

- Looks like you accidentally included a bunch of other unrelated commits here.
- Since you're using `strtolower`, does this make all the initial headings lowercase? We'd probably prefer strtoupper if that's the case.

Other than those little points, this looks good to me.

euanfergusson:
will figure out where the additional commits came from (they arent mine) 

Headings remain uppercase (either way - that should be handled in css) 

screen shot 2017-05-15 at 16 05 15

zerocrates:
Ah sure, it's a text-transform: uppercase there.

Still I think we'd prefer upper vs. lower, even though it's just in the markup.

As for the extra stuff, it's probably just from the branch you were working from... cherry-picking to a new branch directly off develop would be an easy solution to get just the last commit alone.

euanfergusson:
moving this to pr 925

PR #926, opened 2017-05-17
allow the addition of role labels to the ACL

euanfergusson:
the addition of these functions allow roles to be added to the Omeka ACL, making them available to admin users in the admin UI

### DIFF
--- a/application/src/Permissions/Acl.php
+++ b/application/src/Permissions/Acl.php
@@ -109,7 +109,6 @@ class Acl extends ZendAcl
      * @param $roleId
      * @param $roleLabel
      */
-
     public function addRoleLabel($roleId, $roleLabel)
     {
         $this->roleLabels[$roleId] = $roleLabel;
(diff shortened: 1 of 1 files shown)


zerocrates:
An `addRoleLabel` method should probably just take the role ID and the label as separate arguments, and then not do any funny stuff with case changes or string replacement. Similarly `removeRoleLabel` should just take the role ID directly and use it as the key to unset.

euanfergusson:
I'm fine with that, changes made and pushed

zerocrates:
Travis is probably still going to be mad at you when running the code style checker for the blank line at the top of your docblocks.

euanfergusson:
fixed

zerocrates:
You can run `gulp test` locally to see the test results, or `gulp test:cs` to just run the code style checks alone, or in the case of failure check the "X" icon next to your commit in a PR.

In this case, you'll see that the blank line following the docblocks is also a style violation.

PR #933, opened 2017-05-23
Chosen select boxes

jimsafley:


jimsafley:
@mebrett Please confirm that the select boxes work as expected. Check out the `chosen-select-boxes` branch and open the item add/edit page. The template and class select menus should be styled to include a search box that filters the contents in real time, and you should be able to clear a selected option by clicking the adjacent "x". Note that the search box should only appear when there at least 10 options.

mebrett:
Template and Class select menus look different, prompt to select a template/class. Real time updates as I typed, and was able to clear selections by clicking the x.

I can replace the resource using keyboard only but can't clear it completely without clicking (on the previous version I could arrow key up to "select resource" which is basically null)

PR #951, opened 2017-06-20
Bugfix - deleting pages ignores site

euanfergusson:
If there are 2 pages with the same slug on different sites , the delete page functionality currently ignores the site id and deletes the first page created with that slug, not necessarily the one on the site you are working on.


zerocrates:
Well that's an embarrassing one...

Issue #949, opened 2017-06-19
Improve link between media and items


pols12:
I think media should be semantically linked to items with a customizable RDF property.

Otherwise, the resource template should include a way to specify media like one for properties. Indeed, it could be useful to require a medium for a resource and to set a label for the medium.

jimsafley:
These are worthwhile discussions, and something we're thinking about.

> I think media should be semantically linked to items with a customizable RDF property.

One idea we have is to add "Media" to the "Omeka resource" data type ("Item" and "Item set" are already included), but the sidebar select _only_ displays media that is assigned to the item and ingested. This would use our current workflow and would allow more than one property relationship between an item and its media. Omeka would reflect the semantic link in the values list as well as in the JSON-LD serialization of the item

Another idea is to provide a property select for each media in the "Media" tab of the item add/edit page. When a user assigns a property to a media, Omeka would reflect this semantic link in the the item's "Linked resources" tab as well as in the JSON-LD serialization of the item. This would require new functionality and workflow.

Do you have any other ideas about interface?

> Otherwise, the resource template should include a way to specify media like one for properties. Indeed, it could be useful to require a medium for a resource and to set a label for the medium.

This is a good idea, though tough to implement. We'll consider it.

pols12:
Adding "Media" to "Omeka ressource" data type seems to be a good implementation, in my humble opinion. But I think the select panel should propose to ingest new media (such as for theme assets).

In this case, it would be nice if resource templates would be able to precise the resource type (between "Item", "Item Set" and "Media"), when "Resource" would be chosen as data type.

jimsafley:
> In this case, it would be nice if resource templates would be able to precise the resource type (between "Item", "Item Set" and "Media"), when "Resource" would be chosen as data type.

This is a work in progress in #970.

PR #954, opened 2017-06-21
Select media resources as values

jimsafley:
(refs #949)


jimsafley:
@mebrett Testing: the `experimental-media-value` branch adds "Media" to the "Omeka resource" value type. You should _only_ see the "Media" button when editing an item. After selecting "Media", the sidebar should _only_ list media that's assigned to the item. After selecting the media resource, it should show up as normal on the edit form, alongside a media icon. After saving the item, the show page should render the media value as normal, alongside a media icon.

mebrett:
Media option shows up with other Omeka S Resources when editing an item. Does not show up when adding an item as the media hasn't been ingested yet. 

When used, media show up displaying their title as link, with an icon (indicating media type?)
screen shot 2017-06-27 at 1 15 20 pm

kimisgold:
@mebrett The icon indicates the value type, so the 'image' glyph is indicating it is a 'media' value (as opposed to 'item' or 'item-set' values).

mebrett:
@kimisgold thanks for the clarification (I'm sure a user at some point will ask)

PR #956, opened 2017-06-22
Bugfix/nlw 302 trailing slash redirect

euanfergusson:
this is already fixed on apache as part of omeka-s repo. Fixing here for ngnix only.


zerocrates:
PR to wrong place?

euanfergusson:
apologies

Issue #900, opened 2017-05-01
Logged-in indicator


zerocrates:

patrickmj:
If we are going to go with anonymous users being able to switch languages, it seems like the place to make that switch might be very similar to how the logged-in indicator gets stuffed onto the page.

elrossco22:
one to follow @stephenwf @euanfergusson

mebrett:
@kimisgold particular branch?

kimisgold:
* When logged in, a header bar appears above the page layout featuring links to:
  * installation admin
  * site admin
  * user admin
  * logout
* When logged out, the header only appears if the site setting "Always show user bar on public views" is active.

kimisgold:
This is on the `user-bar` branch of omeka-s and the default theme. If all that works, I'll apply the user bar to the rest of the themes for testing.

mebrett:
On branch user-bar and I see the site setting checkbox but there is not a bar across the top on the public views of pages (when logged in and/or with the box checked and logged out)

mebrett:
Seeing it now. It works. 
Some thoughts: 
* the arrow between the install admin and site admin feels like it should present a dropdown. Maybe just a dot instead? 
* Will people think that the edit icon goes to edit the page (which is the behavior on WordPress) rather than the site overall?
<img width="325" alt="screen shot 2017-06-28 at 4 19 13 pm" src="https://user-images.githubusercontent.com/2314070/27658356-bfde6436-5c1d-11e7-8557-218d40e3b973.png">

kimisgold:
The arrow glyph is an easy change. I could also replace the edit icon with more obvious text. It currently goes to the 'site info' edit page, but is there a more useful page to direct administrators torwards?

mebrett:
I feel that because install admin and site admin go to similar places, there might not be a need for the edit button? You can click site admin and get back to the basics for your site, and then go from there to whatever it was you wanted to edit (theme, collecting form, whatever)

PR #965, opened 2017-06-29
Provide a header bar partial to indicate when user is logged in and a site setting to always show user header bar. (#900)

kimisgold:


zerocrates:
Merged, minus the last merge with develop.

PR #966, opened 2017-06-30
Feature/allow full access to config forms

euanfergusson:
Changing the way the module config form is rendered. This approach allows attributes on the form tag itself to be altered in the module config, specifically the 'enctype' attribute to allow module config forms to upload files.

### DIFF
--- a/application/src/Controller/Admin/ModuleController.php
+++ b/application/src/Controller/Admin/ModuleController.php
@@ -277,7 +277,7 @@ class ModuleController extends AbstractActionController
         }
 
         $view = new ViewModel;
-        $view->setVariable('configForm', $moduleObject->getConfigForm($this->viewRenderer));
+        $view->setVariable('configForm', $moduleObject->getConfigForm());
         $view->setVariable('module', $module);
         return $view;
     }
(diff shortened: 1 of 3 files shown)


euanfergusson:
any thoughts on this PR

zerocrates:
Sorry, I thought I had sent a response here.

As you've got it here I think it's problematic, since we have a fair number of modules not using a Form object or using their ability to output arbitrary markup.

Still, I see the need to alter the particulars of the Form in ways we're not currently allowing. I have a couple ideas for how to allow both:

1. Create and pass in a Form object to the method. Omeka only uses it to output the openTag, so the module can set whatever attributes it wants on the form to alter the tag as necessary. This is a little weird, as it means the module could do all sorts of other stuff to the Form that would be ignored (like adding elements), but it would work.
2. Make Omeka accept *either* a string or a Form from this method. If it's a string, act as it does now. If it's a form, just render the form. We'd have to shift things a little so it would do the appropriate changes to the passed-in form.
3. Pass the attributes out-of-band somewhere, like a passed object just for this purpose holding key-value pairs for form tag attributes. I'm not too fond of this option.

PR #968, opened 2017-07-04
change the translator to use zend core

euanfergusson:
This change uses the core zend translator which has additional functionality that it would be useful to allow Omeka to use (particularly the adding of source translation files)


zerocrates:
This change will break some of the translations (those that rely on the Message class).

I assume it would also be fine for your purposes if you could access the "real" Translator underlying the Omeka one? (It's just the Zend one at bottom still)

zerocrates:
We've introduced a method on our translator facade to get to the "real" one: `getDelegatedTranslator()`, which I think will serve your needs as well.

PR #984, opened 2017-07-19
Add filters to resource selector

jimsafley:


zerocrates:
The Filter search header/arrow thing should be an active click area along its entire width

PR #1012, opened 2017-08-03
Allow addon packages to specify a custom install name

garyttierney:
Addon package names are often different to the folder name the module or theme is intended to be installed under.  This allows specifying a custom installation name like in `composer/installers`.


zerocrates:
Looks good to me, thanks!

Issue #1001, opened 2017-07-25
Resource template datatype dropdown disappears in FF when zoomed in


patrickmj:
When zoomed in, the dropdown for selecting a datatype for a property in a resource template disappears.

![datatype](https://user-images.githubusercontent.com/153262/28581159-8f6bd118-712f-11e7-94bd-26c53ed2b754.png)

jimsafley:
This is similar to what I'm seeing in Chrome. Some of the data type select options get hidden under the fold in the sidebar. Oftentimes it's impossible to scroll to the last few options. At that point there are three scrollbars next to each other.

kimisgold:
In Chrome at 125% zoom, I see a scrollbar for page content and a scrollbar for sidebar content—Firefox only scrolls the page content, so I'll look into why. 

@jimsafley Theoretically, the chosen select plugin creates extra room at the bottom of select's container so that the dropdown menu is visible. Another developer offered [this modification](https://github.com/harvesthq/chosen/issues/155#issuecomment-173238083) on top of chosen core that I think allows the menu to appear above the select—if you think this looks promising, I can help you style it if you implement the JS.

PR #1013, opened 2017-08-03
Chosen layout fix

jimsafley:
See #1001


zerocrates:
I pulled this in but on closer inspection it causes some problems:

- If the viewport is fairly short, it will try to "drop up" even selects at the top of the screen, causing them to clip off the top
- After the first "drop up", the menu sits in the markup below the select and takes up space (giving you extra scroll distance in the sidebar). If you scroll into that space, you get weird results because the menu moves when it goes to drop up, so your pointer is sitting over the list and will immediately select the entry under it if you mouseup.

zerocrates:
@kimisgold the problem with the "reserves extra space" thing might be solvable within our CSS: we can add `overflow: hidden` to .chosen-container, and also `overflow: visible` to .chosen-container.chosen-with-drop

We can probably tweak the script to prevent "drop up" if there's not enough space above the element to fit the menu as well.

#1001 wasn't really Chosen-specific though, instead a problem with the styling of the sidebar and the button, so it's possible that reversion is a reasonable option as well.

zerocrates:
@kimisgold @jimsafley We need to decide what to do here: either try to do some fixes for these problems, or just back out this change.

I _believe_ that the other work on #1001 means that backing this out won't cause a problem (but it will still cause the resource template sidebar to grow vertically when the Chosen select is opened).

kimisgold:
I'm fine with the simplest solution of reverting, unless @jimsafley thinks the fixes to the dropup are comparably minimal to implement.

PR #1022, opened 2017-08-11
Allow user to remove every value in batch edit and advanced search forms, including the first/only.

kimisgold:


application/asset/js/global.js
```diff
@@ -413,12 +413,20 @@ var Omeka = {
             $(selectedActionClass).addClass('active');
         });
 
+        // Set up multi-value templates
+
+        $('.multi-value.field').each(function() {
+            var field = $(this);
+            var value = field.find('.value').first().clone();
+            field.attr('data-field-template', value.wrap('<div></div>').parent().html());
```
zerocrates: You can set with `.data` just like you get with it, no need to do an end-around through `.attr`

zerocrates:
I envisioned this (the template system) as being something that would take place in the markup, so I'll need a little time to look this different method over.

One thing that seems to be an issue is that you're doing this two ways simultaneously: one using the template-in-data approach in global.js, and another in the advanced search js that just relies on keeping around a jQuery match from load time and cloning off of that. I can't think of any justification for doing both.

PR #1038, opened 2017-08-27
Managed a fallback for assets.

Daniel-KM:
In order to give more flexibilty to theme builders, I added a way to use a theme asset if it exists, else to use the module one. This is similar to the view stack behaviour, but for assets.

### DIFF
--- a/application/src/View/Helper/AssetUrl.php
+++ b/application/src/View/Helper/AssetUrl.php
@@ -27,7 +27,7 @@ class AssetUrl extends AbstractHelper
     /**
      * Construct the helper.
      *
-     * @param string|null $currentTheme
+     * @param Theme|null $currentTheme
      * @param array $modules
      */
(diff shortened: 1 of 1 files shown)


zerocrates:
So if I'm reading this correctly, it's opt-in, since you have to pass true for the third argument, and if you do it will check the current theme to see if the asset exists, then if it does it will use the theme version, otherwise it will just keep current behavior.

Part of the goal with the new assets system was to remove the many checks against the filesystem that the old system did for every asset to determine where to load it from. But, the opt-in nature of this makes that not much of a concern, I think.

My only quibble is over the name. "Fallback" doesn't sound quite right for what this enables. It's more of a theme override than a fallback, in my mind.

Daniel-KM:
The main purpose is to let themes builders a little more freedom, in particular when a asset is not up to date or is customizable.

zerocrates:
Okay, I think we're on the same page.

PR #1040, opened 2017-08-30
Restyle resource value inputs.

### LINKED ISSUE #1024: Reduce value inputs' vertical space

### LINKED ISSUE #1026: Longer labels break in the external link UI
uri-label

We could have these (with flexbox!) just lay out the textboxes to take up whatever space is left after the label, rather than being aligned to each other as they are now. Can't really think of another way to account nicely for long translations here. Of course we can just tack on a little more to the width just to cover this specific case but this just happens to be one translation.

kimisgold:
Restyle resource value inputs to take up less vertical height and to better accommodate varying uri label string lengths amongst languages. (Fix #1024 and fix #1026.)


zerocrates:
What does this do, if anything, on the Sites page edit? I believe they share the basic "top gray bar with trashcan" style for blocks.

kimisgold:
@zerocrates It doesn't do anything. It only targets value field styles on the resource forms. On Sites page edit, the gray bar seems more useful since it includes both the sortable handle and delete button, rather than solely the latter. Block titles are also more informative than value data type titles.

PR #1047, opened 2017-09-06
Fixed "finialize" one more time.

Daniel-KM:


zerocrates:
We just _really_ wanted to put those finials in.

PR #1048, opened 2017-09-06
Replaced the key by the true property id.

Daniel-KM:
I'm currently working on integrating features of CsvImport plus in CsvImport, in particular the update process, so this modification may simplify the process and, anyway, is more flexible.


Daniel-KM:
There is no test cases for this in Omeka , neither in CsvImport, so this is not a patch to merge currently.

PR #1091, opened 2017-10-16
Feature/module weighting

Daniel-KM:
A simplification of #904.

I have the same problem with a module [Group](https://github.com/Daniel-KM/Omeka-S-module-Group) (to be published), where I need to attach events before some other modules. I can't manipulate the config via an init in the module, because there is no configuration. And there is no dependencies.

So I found two solutions. The simplest one is to set the module in `application/config/application.config.php`, at the end of the array "modules", just after "Omeka". It works fine, but requires to change a default setting. 

The second is to order modules. Currently, they are ordered according to the file system timestamp or inodes. At least, they should be ordered alphabetically, but to sort via a weight is also simple. So see commit.

In fact, I use a third solution currently (I detach the first listener), but I can't be totally sure that I remove the good one if another module manipulate it before (this is very rare).

### DIFF
--- a/application/data/install/schema.sql
+++ b/application/data/install/schema.sql
@@ -84,6 +84,7 @@ CREATE TABLE `module` (
   `id` varchar(190) COLLATE utf8mb4_unicode_ci NOT NULL,
   `is_active` tinyint(1) NOT NULL,
   `version` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `weight` tinyint(3) UNSIGNED NOT NULL DEFAULT 0,
   PRIMARY KEY (`id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
 CREATE TABLE `password_creation` (
--- a/application/data/install/schema.sql
+++ b/application/data/install/schema.sql
@@ -84,7 +84,6 @@ CREATE TABLE `module` (
   `id` varchar(190) COLLATE utf8mb4_unicode_ci NOT NULL,
   `is_active` tinyint(1) NOT NULL,
   `version` varchar(255) COLLATE utf8mb4_unicode_ci NOT NULL,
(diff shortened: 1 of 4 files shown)


elrossco22:
👍

zerocrates:
> In fact, I use a third solution currently (I detach the first listener), but I can't be totally sure that I remove the good one if another module manipulate it before (this is very rare).

The event manager itself has a weighting system built in. Is it not sufficient for your needs?

Daniel-KM:
Yes, this is enough for this case, because I don't know any other module that manipulate default listeners.

Daniel-KM:
And the issue may appear if multiple plugins use the events weighting system with the same priority.

zerocrates:
> And the issue may appear if multiple plugins use the events weighting system with the same priority.

Doesn't the same problem occur with any weighting system (for example, this one)? Or are you thinking there's potentially some problem with what the event manager itself does. I'm pretty sure at least that unlike Classic's home-grown event system there's no danger of clobbering events within the manager itself due to having the same weight (after all, everything _not_ picking a weight now, the vast majority, is currently sharing).

Daniel-KM:
This system has no repercussion on the core, and it's easy to change the priority of a module when needed. Perhaps I will require it for a future module.

Daniel-KM:
I simplified the weighting system one more time. Now, the module are managed in an alphabetic order. 

This is important for the previous reasons, plus two points: first, the fact that the modules are displayed in alphabetic order in the main admin board, and this important for the user experience; second, the fact that when two installations of Omeka S are exactly the same, the internal order of the config and the attached events are exactly the same, because they are no more based on file timestamps, so it is predictable, and reduce the possibility of a mysterious bug related to this point that will be hard to identify. The solution 1 can always be used when a module should be loaded first.

Daniel-KM:
See a way to create an issue with #1123.

Daniel-KM:
Thanks!

PR #1100, opened 2017-10-23
Feature/batch edit factorize and trigger events for modules 

Daniel-KM:
It factorizes batch edit item and batch edit item set, and added triggers `view.batch_edit.before / after` (may be renamed) in the view and `form.add_elements / add_input_filters` in the form. 

The main purpose is to simplify the maintenance of the batch-edit (but the resource controllers are not factorized), and to allow modules to set values to add/remove too.

Note that there is an issue for modules : currently, they can't use the event `api.update.pre/post`, that is used by the `AbstractEntityAdapter::batchUpdate()`, because the request is strictly filtered in [application/src/Api/Adapter/AbstractEntityAdapter.php#L375](https://github.com/omeka/omeka-s/blob/develop/application/src/Api/Adapter/AbstractEntityAdapter.php#L375) (default to empty array), so everything in the request is removed except core values of the resources. 

This filter is not used for other resources, so I was wondering if there is a particular reason for that. If not, the filter should be improved to allow a preprocess, but without removing unidentified keys from the raw data (they belong to modules that use the batch-edit).

### DIFF
--- a/application/config/module.config.php
+++ b/application/config/module.config.php
@@ -442,6 +442,7 @@ return [
         ],
         'factories' => [
             'Omeka\Form\ResourceForm' => Service\Form\ResourceFormFactory::class,
+            'Omeka\Form\ResourceBatchUpdateForm' => Service\Form\ResourceBatchUpdateFormFactory::class,
             'Omeka\Form\UserForm' => Service\Form\UserFormFactory::class,
             'Omeka\Form\SettingForm' => Service\Form\SettingFormFactory::class,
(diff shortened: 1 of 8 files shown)


Daniel-KM:
Rebased.

Daniel-KM:
Rebased and simplified, so the issue above is fixed.

Daniel-KM:
Rebased.

Daniel-KM:
Rebased.

Daniel-KM:
Merged via #1101.

PR #1103, opened 2017-10-24
Allow importing of resources that have multiple types

jmiguelv:
When the document being imported contains a class and a property with the same name only one of them was getting imported. That was because of `EasyRdf` `resource.type()` only returns a string with one of the multiple types.

The solution here uses `resource.types()` which returns an array of all the types of a resource and intersects it with `classTypes` and `propertyTypes`.


zerocrates:
Can you give an example of where this is happening in practice?

zerocrates:
As in, a vocabulary document where the same resource is both a property and a class. I'm having trouble wrapping my head around a situation where that could be valid.

jmiguelv:
A vocabulary can have a Class and a Property that have the same name, for example:
```<owl:Class rdf:ID="subject">
...
</owl:Class>
...
<rdf:Property rdf:ID="subject" />
```
When looping through the resources parsed via EasyRdf it gets the following for subject:
```http://ead3.archivists.org/schema/#subject (EasyRdf_Resource)
→ rdf:type → owl:Class, rdf:Property
```

zerocrates:
What I'm looking for is a real-world example file that does this. To my mind, what you and the EasyRDF output are describing isn't a class and property with the same name, they're sharing their entire URL, making them the same resource. It seems to me that this situation would be invalid, or at the very least strongly disfavored, to the exent that I'm not sure we should cater to the possibility.

brian-maher:
Hi @zerocrates 

If you import the file here https://github.com/kingsdigitallab/gpp/blob/develop/metadata/rdf/ead3.rdf

You will find that only ~80 of the entries are imported. I have attached a dump of the graph [here](https://github.com/omeka/omeka-s/files/1415674/easyrdf_ead.html.txt)
; you will see that some are both classes and properties. Because of the implementation of the type() method:

`The type will be a shortened URI as a string. If the graph has multiple types then the type returned may be arbitrary. This method will return null if the resource has no type.`

In the case of the above file, this results in a large number of properties being missed, because the type() method is returning class. Checking for an intersect with the types() array will ensure these properties are not missed, and means the import isn't reliant on a function which, by definition, is arbitrary.

PR #1107, opened 2017-10-25
Set the site option "show_user_bar" ternary option.

Daniel-KM:
I added a new value to the site setting `show_user_bar`, so it is possible now to display the user bar "never", "when identified" or "always", and not only "when identified" or "always".


Daniel-KM:
Rebased.

Daniel-KM:
Rebased.

PR #1108, opened 2017-10-25
Convert the user bar into a helper

Daniel-KM:
Based on the previous pull request, I converted the user bar into a helper, so it is possible to add new tools in it. In fact, there are two ways to add new tools: a helper, so it's possible to upgrade the https://github.com/omeka/plugin-EditLink, or a trigger, or even a block, but it's simpler via a helper and modules for now.

Note that this change requires an update to the official themes, since the user bar is now just `echo $this->userBar();` (the partial is still available, but it requires a value for "user" and checks are not done). I can push this update if needed.

### DIFF
--- a/application/config/module.config.php
+++ b/application/config/module.config.php
@@ -391,6 +391,7 @@ return [
             'themeSettingAssetUrl' => View\Helper\ThemeSettingAssetUrl::class,
             'formColorPicker' => Form\View\Helper\FormColorPicker::class,
             'thumbnail' => View\Helper\Thumbnail::class,
+            'userBar' => View\Helper\UserBar::class,
         ],
         'factories' => [
             'api' => Service\ViewHelper\ApiFactory::class,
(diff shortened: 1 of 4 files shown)


Daniel-KM:
The last commit avoids to update old themes, but the check for the user bar should be done when the helper is not used.

zerocrates:
I like this change, but I think we're going to wait until post-1.0 for it.

Daniel-KM:
Rebased.

Daniel-KM:
Rebased.

zerocrates:
I'm understanding the change correctly that the bar will render with no theme changes, but switching over to use the helper is required to get the always/never/logged-in switch to work, yes?

Daniel-KM:
Last year, the idea was to avoid the upgrade of themes, so the upgrade works with any theme, and that's why the radio uses values "-1", "0", and "1", like a checkbox. These values can be changed if you want to require a change in themes.

PR #1115, opened 2017-10-27
Added attributes to linkPretty.

Daniel-KM:
I added the possibility to add attributes to `linkPretty()`, like in `link()` and `linkRaw()`. 

I was wondering too if the default class `resource-link` should be added in all cases or not.


zerocrates:
Yeah, I think if we're going to allow custom attributes here, we want to force `resource-link` into the class at all times, since it's the backbone of the styling.

Daniel-KM:
Ok, I add it.

zerocrates:
Looks good to me; should allow this to be used in a couple situations it otherwise wouldn't work for.

PR #1123, opened 2017-10-30
Displayed modules alphabetically in the left admin board.

Daniel-KM:
The list of modules in the left admin bar is displayed in various order, depending on file system, and on the order the modules are enabled, so it is not ergonomic. It is simpler to set a main order for all modules, so the same order is used anywhere.

### DIFF
--- a/application/src/Module/Manager.php
+++ b/application/src/Module/Manager.php
@@ -120,6 +120,14 @@ class Manager implements ResourceInterface
     }
 
     /**
+     * Sort modules alphabetically.
+     */
+    public function sortModules()
+    {
+        ksort($this->modules);
+    }
+
+    /**
      * Get all registered modules
      *
      * @return array
--- a/application/src/Service/ModuleManagerFactory.php
+++ b/application/src/Service/ModuleManagerFactory.php
@@ -141,6 +141,9 @@ class ModuleManagerFactory implements FactoryInterface
             }
         }
 
+        // Reorder modules.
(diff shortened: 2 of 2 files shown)


zerocrates:
Should I presume that this is basically alternate to your "weighting" PR, which I seem to remember also had alpha sorting as a default?

Daniel-KM:
The code is the same, but the aim is different.

Daniel-KM:
Rebased.

Daniel-KM:
I found a way to create a hard-to-find issue on this point. Create two modules that uses a view:

```php
    'view_manager' => [
        'template_path_stack' => [
            dirname(__DIR__) . '/view',
        ],
    ],
```

Then create two common views named `view/common/resource-sidebar.phtml`. Write anything in them:

```html
<div>Module A</div>
```

and 

```html
<div>Module B</div>
```

Then, create two identical installs, with the same modules and the same database, but on a different system and a different order of install. So the default order of the module will be `module A` then `module B` in the first case, and `module B` them `module A` in the second case. 

In the first case, the view Module B will be displayed; and in the second case the view Module A will be displayed.

So the only difference is the different order of modules, due to a different file system or a different order of install of modules. So a very hard to find issue. The commit won't change the issue, but it makes it reproductible on any system and any install order, so easier to fix.

Daniel-KM:
Rebased.

Daniel-KM:
Rebased on 1.3.

PR #1132, opened 2017-11-06
Fix/rights browse

Daniel-KM:
Cleaned batch `update` / `batch delete` / `update` / `delete` rights in item/browse and item-set/browse.


Daniel-KM:
It is interesting to allow every people to batch process something, even if the user has no rights to batch edit or batch delete: for example, a user may be able to export a selection of items (see the module Zotero in pr https://github.com/omeka-s-modules/ZoteroImport/pull/21). So the last commit fix this point.

Daniel-KM:
Rebased.

Daniel-KM:
Rebased on 1.3. Available via module [Next](https://github.com/Daniel-KM/Omeka-S-module-next).

Daniel-KM:
Rebased on 1.4. Available via module [Next](https://github.com/Daniel-KM/Omeka-S-module-next).

PR #1140, opened 2017-11-13
Fix/message i18n

Daniel-KM:
A missing `use Omeka\Stdlib\Message;` in application/src/Site/BlockLayout/Fallback.php.


zerocrates:
This is pulling in one of Patrick's prior commits, which we squash-merged.

Daniel-KM:
I rebased it.

PR #1145, opened 2017-11-16
Upgrade PHPUnit to v5 in way compatible with v6

keradus:
Brought to you by [PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)

Upgrade PHPUnit to v5 in way compatible with v6.
That means, you could start using PHPUnit 6 as well without any changes in tests, as soon as you fulfill PHPUnit 6 requirements (bumping php itself to v7)


zerocrates:
Thanks, this is very helpful.

PR #1151, opened 2017-11-20
Fixed doc block.

Daniel-KM:


Daniel-KM:
I rebased it without it.

Issue #1146, opened 2017-11-16
New addon version notification


zerocrates:
Separating from #1144

zerocrates:
My initial thought here is that client-side is the way to go. We'll get etag-based caching for free already and could trivially add a one-day Expires to the addons jsons on the omeka.org side of things to be more aggressive if we wanted.

PR #1160, opened 2017-11-28
Version notification

### LINKED ISSUE #1144: New version notification
A la Classic.

We could probably either hook it to Github's "latest release" API, or use the same system Classic uses where we host an exceedingly simple API just declaring the latest released version.

jimsafley:
fixes #1144 #1146


zerocrates:
Should use HTTPS for the addons JSONs as well.

zerocrates:
We probably should lose the "Click here" in all those messages as we don't really use that pattern elsewhere. The main S check should probably just say "get the new version" too, since it's really pointing at the download page and not "instructions" per se.

jimsafley:
I suppose the notifications need styling? @kimisgold

zerocrates:
Yes, I was thinking so.

PR #1161, opened 2017-11-29
Add option to add new page to navigation

### LINKED ISSUE #1117: Consider an option on page add to automatically add it to the navigation
When creating a new page for a site, it might be an additional convenience to have a checkbox to automatically add it to the navigation. Especially for the earliest stages of building a site, that might be a handy convenience.

jimsafley:
fix #1117


patrickmj:
This seems like possibly a difficult check, but when editing a page that's already in the navigation it might be helpful to signal that. Currently it looks like it adds the checkbox regardless.

jimsafley:
@patrickmj That reminds me to remove the "Add to navigation" checkbox from page edit.

zerocrates:
I think having it only on Add is the sanest option. The original issue was in terms of "new" pages and that keeps it nice and simple.

PR #1164, opened 2017-12-02
Centering media in page block

jonfackrell:
Added option to center media in page block.

Template designers must add some css like the following:

.center.file {
    margin: 0 auto;
}


zerocrates:
A "center" alignment option is in 2.1.0 and up, along with a feature so we can provide the base styling for it and other block options without updating themes.

PR #1165, opened 2017-12-04
Fix/install php

Daniel-KM:
There is an error when installing on a php 5.6 server : random_bytes is not present, so there is a fatal error. 

Note that the [workaround](https://secure.php.net/manual/en/intro.csprng.php) for this php version is to use the library [paragonie/random_compat](https://github.com/paragonie/random_compat), but it is marked 

> Although this library has been examined by some security experts in the PHP community, there will always be a chance that we overlooked something. Please ask your favorite trusted hackers to hammer it for implementation errors and bugs before even thinking about deploying it in production.

So Omeka S should not be marked available with php 5.6, unless you warn the user of it.

### DIFF
--- a/application/src/Installation/Task/CheckEnvironmentTask.php
+++ b/application/src/Installation/Task/CheckEnvironmentTask.php
@@ -43,10 +43,14 @@ class CheckEnvironmentTask implements TaskInterface
      */
     protected function testRandomGeneration()
     {
-        try {
-            random_bytes(32);
-        } catch (\Exception $e) {
-            $installer->addError('Omeka is unable to securely generate random numbers.');
+        if (function_exists('random_bytes')) {
+            try {
+                random_bytes(32);
+            } catch (\Exception $e) {
+                $installer->addError('Omeka is unable to securely generate random numbers.');
+            }
+        } else {
+            $installer->addError('Omeka is unable to securely generate random numbers. Install random compat (https://packagist.org/packages/paragonie/random_compat).');
         }
     }
 }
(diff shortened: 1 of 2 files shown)


zerocrates:
paragonie/random_compat is already installed by Composer, as a dependency of zend-math (note that it appears in composer.lock currently). I don't disagree with having it be explicitly required directly by us though.

I don't know about the change to the installer's check though... Since it's installed by Composer I don't think there's any more call to have a message specifically about it than there is to have one for any of the many other packages that are required.

I'm not sure how you'd be getting the error you describe, unless you've changed the dependencies.

Daniel-KM:
So the fix to add the library is useless (but the fatal error is real). 

In fact, I don't find the issue myself, but somebody with a php 5.6 server with Omeka Classic reported it to me. So I go back to him.

zerocrates:
Hmmm, okay.

We've definitely run S on a variety of 5.6 servers without this issue cropping up. My brief perusal of the zipped download doesn't reveal any obvious problem: random_compat is included under vendor/ and the autoloader seems to be set up to load it as it should as well.

Possibly an error in transfering the files to the server?

Daniel-KM:
So I rebased and removed the paragonie/random_compat, but let the other commits. There is an issue with php 7.2.

zerocrates:
Yeah, I saw the zend-form issue.

I think they just released the updated version that should fix it.

zerocrates:
I think I'm going to just add 7.2 to the test matrix separate from this PR... I'm not totally sold on the wisdom of checking for the function when it should be guaranteed to exist, at least not without more understanding of the situation that's actually causing the problem.

Daniel-KM:
Deprecated.

PR #1168, opened 2017-12-07
Allowed to use a select element in theme form.

Daniel-KM:
When a `theme.ini` contains a select form, an empty value is not allowed. This fixes it. 
In fact, the Select doesn't accept a default value nowhere, but this is a Zend element, so it may be changed globally via getInputSpecification or something else, but for now this is enough.

### DIFF
--- a/application/src/Controller/SiteAdmin/IndexController.php
+++ b/application/src/Controller/SiteAdmin/IndexController.php
@@ -355,20 +355,24 @@ class IndexController extends AbstractActionController
             $form->setData($oldSettings);
         }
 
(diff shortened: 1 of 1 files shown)


zerocrates:
This is one of these areas where I can't quite figure why the default is the way it is, but ah well. It's not an uncommon sentiment I have about the form component.

Does this work using `allow_empty` true rather than `required` false?

My preference would be to apply this only to selects and only ones that have an empty option set, but I'm not sure how much it really matters to do so.

Daniel-KM:
Yes, it works with `allow_empty` only.

PR #1172, opened 2017-12-13
Fix/docblocks

Daniel-KM:
@inheritDoc is useful only when there is a specific doc.


zerocrates:
Yeah, we picked this up as it was basically Zend Framework house style at the time. I'm not sure if they've dropped it, but it is fairly silly to have all over the place by itself.

Daniel-KM:
Merged #1291.

PR #1177, opened 2017-12-19
IIIF image URL is mandatory

dodeeric:
Add the red star in front of the input field to indicate it is mandatory.


dodeeric:
Travis-CI error. Have to check.

dodeeric:
To be checked.

PR #1178, opened 2017-12-21
Make IIIF image URL mandatory

dodeeric:
Make the "IIIF image URL" field mandatory (it will also add a red star in front of the input field).


zerocrates:
Shouldn't the attribute be `required`, not `require`?

dodeeric:
You are right! I have fixed the typo.

PR #1179, opened 2017-12-22
Fixed check of local file for asset url.

### LINKED ISSUE #1147: Cachebusting for assets
This was (not too long ago) contributed to Classic and there's no good reason we shouldn't have it for S.

In the pre-release cycle we were a little more predisposed to not care about caching-based problems with JS and CSS, but as we're talking about an increased userbase avoiding this kind of stuff is more important.

For the core simply the version string in a query string should suffice. We could also use the versions from modules and themes... though we'd have to do a little more legwork making those available. Since the asset system already draws lines between those sources it should make things a little easier.

Attention probably needs to be paid to the override feature that was recently added for assets as well.

Daniel-KM:
Fix #1163.


Falteri:
After that fix it looks like some images are missing, e.g. http://<site>/application/asset/vendor/openseadragon/images/?v=1.1.0-alphazoomin_hover.png
But maybe there is some other problem on my site?

PR #1182, opened 2018-01-08
Feature/log writers

Daniel-KM:
When I build a module to manage logs directly in the admin board in particular ([Daniel-KM/Omeka-S-module-Log](https://github.com/Daniel-KM/Omeka-S-module-Log)), I see that the config or Omeka S doesn't follow some points of a standard Zend config. This is the case for the logger, so I updated the config for it, partially.

### DIFF
--- a/application/config/module.config.php
+++ b/application/config/module.config.php
@@ -177,7 +177,7 @@ return [
     ],
     'logger' => [
         'log' => false,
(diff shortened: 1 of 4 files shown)


Daniel-KM:
Fully managed in module [Log](https://github.com/Daniel-KM/Omeka-S-module-log).

PR #1185, opened 2018-01-10
Advanced search options

jmiguelv:
Limits advanced search, search by class and search by value options, to resource classes and properties that are being used by items.


zerocrates:
I'm concerned about the scaling/performance impacts here: you're doing an extra API query for each property. Pulling in things like schema can quickly get you to lots of properties, and each of those queries is going to get slower as the set of resources grows larger, too.

As a side issue, you're also doing the query even if the selector is set to show everything, which is just unnecessary.

Probably the best option for doing this would be to do it as a join, meaning implemented as a "search" option in the property and class API adapters.

The feature itself is definitely one I think makes sense, and that we could hook up to global and/or site settings.

zerocrates:
Closing in favor of #1793

PR #1189, opened 2018-01-18
Skip bad queries in AbstractResourceEntityAdapter (instead of throwing an exception)

pols12:
Fix `continue` use which has to continue the `foreach` loop, not only the `switch`, when a nonexistent query type is used.


zerocrates:
looks good to me

Issue #1193, opened 2018-01-22
Omeka S dies when trying to add too-large IIIF image


crism:
When importing an IIIF image, _e.g._ from [here](https://iiif.library.illinois.edu/dls/iiif/2/7e64fa40-0b13-0134-1d55-0050569601ca-2/info.json), Omeka S [constructs a URL](https://github.com/omeka/omeka-s/blob/develop/application/src/Media/Ingester/IIIF.php#L84). If the image at that URL is larger than the allowed maximum, the IIIF server will [return a 403 Forbidden response](https://iiif.library.illinois.edu/dls/iiif/2/7e64fa40-0b13-0134-1d55-0050569601ca-2/full/full/0/default.jpg). Omeka S does not check for this, and when it tries to [delete the temp file](https://github.com/omeka/omeka-s/blob/develop/application/src/Media/Ingester/IIIF.php#L97), it crashes.

bencomp:
I may be missing something here, but one of the features of the IIIF Image API is that you can request images in smaller sizes, for example to use as thumbnails. So instead of downloading a full-size image and resizing it on the side of Omeka S, I would recommend to download a thumbnail-sized image. Or is that what you were saying, @zerocrates?

zerocrates:
So, the problem for us here is just that IIIF image servers can support various subsets of the spec features. The full/max size is (supposedly) guaranteed to be supported even by the lowest levels of compliance with the spec, so that's what we use.

Obviously there's the idea that we could use the IIIF features themselves to get the size we want directly, but that resizing is not guaranteed to be available either.

Some version of this that depended on the `sizes` to be present, or looked at what features the server advertises support for to then try to use them, could be possible, but I'm not sure either provides a whole lot of practical benefit.

PR #1194, opened 2018-01-22
omeka/omeka-s#1193: quick patch to prevent crashing on too-large image

crism:
This produces slightly weird behavior, as the full-size image fails to import, but it doesn’t crash and the IIIF viewer seems to work correctly.


zerocrates:
Seems like a reasonable first step to me

PR #1205, opened 2018-01-29
Updated readme.

Daniel-KM:


zerocrates:
My laziness has won out again, as this content has long since moved from the README to the documentation.

PR #1207, opened 2018-02-01
Allow to give attributes to img tag in media show page

pols12:
Especially `alt` attribute which is recommanded for accessibility.

[Yanyufish asked to have alt tag for images on the forum](https://forum.omeka.org/t/no-alt-tag-for-images/5449).

### DIFF
--- a/application/src/Media/FileRenderer/ThumbnailRenderer.php
+++ b/application/src/Media/FileRenderer/ThumbnailRenderer.php
@@ -11,7 +11,8 @@ class ThumbnailRenderer implements RendererInterface
     ) {
         $thumbnailType = isset($options['thumbnailType']) ? $options['thumbnailType'] : 'large';
         $link = array_key_exists('link', $options) ? $options['link'] : 'original';
-        $img = $view->thumbnail($media, $thumbnailType);
+        $attribs = isset($options['attribs']) ? $options['attribs'] : [];
+        $img = $view->thumbnail($media, $thumbnailType, $attribs);
         if (!$link) {
             return $img;
         }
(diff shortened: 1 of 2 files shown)


zerocrates:
Passing attributes seems like a good option to have here, though we might want to tweak the option name to clarify that these are attributes for an image (or even thumbnail image) only.

The specific use of the alt for the media on its own show page, I'm not so sure of though. In that context the title is already on the page and highly accessible through the title and heading elements. I'm not sure that having it also as the alt of the image is actually an accessibility improvement over the default blank alt there.

pols12:
> The specific use of the alt for the media on its own show page, I'm not so sure of though. In that context the title is already on the page and highly accessible through the title and heading elements. I'm not sure that having it also as the alt of the image is actually an accessibility improvement over the default blank alt there.

Indeed, title brings no new information, that’s what @patrickmj has said [on the forum](https://forum.omeka.org/t/no-alt-tag-for-images/5449/4) too.

I proposed to use `dcterms:description`. I’m not a semantic web specialist: would it be interesting to use a custom property (in `omeka` vocabulary) which would be defined as `owl:equivalentProperty` of [`cwork:altText`](http://www.bbc.co.uk/ontologies/creativework/altText)?

zerocrates:
A custom or otherwise set aside property or column just for alt text might
well be the best answer here. I think there's likely to be a mismatch
between Description and alt often too.

Of course that's a larger change, touching the data model and the UI.

zerocrates:
I cherry-picked the first commit out of here (just adding the attribute passing for thumbnail rendering), but I changed the option name to thumbnailAttribs to avoid any future conflict or confusion with other possible attribute options.

PR #1210, opened 2018-02-07
Allow item set sorting

pols12:
In ItemSetAdapter, moves ORDER BY site_item_set.position from `buildQuery()` to end of `sortQuery()`, in order to prioritize user sorting choice.

Note that `$site` seems to be unused.

### DIFF
--- a/application/src/Api/Adapter/ItemSetAdapter.php
+++ b/application/src/Api/Adapter/ItemSetAdapter.php
@@ -86,7 +86,7 @@ class ItemSetAdapter extends AbstractResourceEntityAdapter
             }
         }
 
-		if (!empty($query['site_id'])) {
+        if (!empty($query['site_id'])) {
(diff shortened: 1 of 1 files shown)


zerocrates:
The issue with doing that is that the innerJoin affects the actual returned records from the query. We probably want to keep the sort methods just handling sorting. But, moving the default position sort there is probably the way to go in terms of fixing this problem.

As for site not being used... that is interesting for sure... It actually seems like there's a(nother) fairly significant bug there, if I'm reading correctly.

zerocrates:
Well, I guess it's not a bug per se... there's just an oddity where the passed-in ID is directly used in the query to limit to the correct site, where it could be read from the actually-retrieved site record... this section should either be actually checking against $site and aborting if its not found, or else dropping the check and just using the ID directly with no read at all. We probably want to go the former route...

PR #1214, opened 2018-02-23
Add filtering form view helper for browse pages

jimsafley:
Needs styling.


zerocrates:
What's the rationale behind the hidden and copying the value from the select to there? Doesn't seem like it would really add anything beyond just having the select directly...

zerocrates:
So, the position is that we'd prefer to have this as a dropdown rather than links? Because it seems to me that the javascript here is basically there to make these act more like they would as a series of links.

@kimisgold?

PR #1215, opened 2018-02-23
Triggered event "user.logout" when logout.

Daniel-KM:
When an external app is used to identify users, it should be warned that the user logs out.

### DIFF
--- a/application/src/Controller/LoginController.php
+++ b/application/src/Controller/LoginController.php
@@ -75,8 +75,14 @@ class LoginController extends AbstractActionController
     public function logoutAction()
     {
         $this->auth->clearIdentity();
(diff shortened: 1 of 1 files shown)


zerocrates:
Looks fine to me.

Out of curiosity, is there something you're working on that would use this?

Daniel-KM:
For a project, we build a saml module and there is another module with a js library. They work fine, but they are not published, because the choice of the license is not yet resolved.

zerocrates:
SAML support is something a lot of institutions will be interested in, I'm sure.

Daniel-KM:
Sadly, I'm not the one who decides on this project.

PR #1223, opened 2018-03-13
Handle and log errors thrown during render

garyttierney:
Omeka was previously only listening on exception events that came from the
`EVENT_DISPATCH_ERROR` event.  However, errors can also be signalled as
`EVENT_RENDER_ERROR` events, which will never be logged with the current
implementation.

### DIFF
--- a/application/src/Mvc/ExceptionListener.php
+++ b/application/src/Mvc/ExceptionListener.php
@@ -24,6 +24,7 @@ class ExceptionListener extends AbstractListenerAggregate
     public function attach(EventManagerInterface $events, $priority = 1)
     {
         $this->listeners[] = $events->attach(MvcEvent::EVENT_DISPATCH_ERROR, [$this, 'handleException'], -5000);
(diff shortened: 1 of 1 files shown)


zerocrates:
Yeah, this isn't as obvious at first glance since our "custom" exception handler is designed to not really change what exceptions look like and how they're rendered, but you're right that leaving out RENDER_ERROR means those don't get _logged_... and we definitely want to maintain the expectation from Classic that whenever you see the "encountered an error" page, it's being logged.

Alternatively it might be cleaner to separate out just the logging and have only _that_ listening to both error types... The way it is now in this PR, there'd be a small side effect: uncaught exceptions from API requests originating from the views (the API helper) will show as 404 and 403 errors if they're "not found" or "permission denied" exceptions, respectively. I'm a little worried that would be confusing. Right now those errors are displayed as generic exceptions when they occur if they're coming from the view, which is how I think it makes the most sense.

zerocrates:
This was implemented as 4aeca555ed in a slightly different form that accounted for my earlier concerns about incorrectly showing the 404/403 error page.

PR #1243, opened 2018-04-12
Store original file size

### LINKED ISSUE #1242: Store original file size during ingestion
It would be helpful to store file size in case something needs to act on it (e.g. sorting, filtering, conditionals).

jimsafley:
fix #1242

### DIFF
--- a/application/view/omeka/admin/media/show-details.phtml
+++ b/application/view/omeka/admin/media/show-details.phtml
@@ -25,6 +25,13 @@ $item = $resource->item();
     </div>
     <?php endif; ?>
 
+    <?php if (is_numeric($resource->size())): ?>
+    <div class="meta-group">
+        <h4><?php echo $this->translate('Size'); ?></h4>
+        <div class="value"><?php echo sprintf('%s bytes', number_format($resource->size())); ?></div>
+    </div>
+    <?php endif; ?>
+
     <div class="meta-group">
         <h4><?php echo $escape($this->translate('Visibility')); ?></h4>
         <div class="value"><?php echo ($resource->isPublic()) ? $this->translate('Public') : $this->translate('Private'); ?></div>
--- a/application/view/omeka/admin/media/show-details.phtml
+++ b/application/view/omeka/admin/media/show-details.phtml
@@ -28,7 +28,7 @@ $item = $resource->item();
     <?php if (is_numeric($resource->size())): ?>
     <div class="meta-group">
         <h4><?php echo $this->translate('Size'); ?></h4>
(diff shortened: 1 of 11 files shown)


application/view/omeka/admin/media/show.phtml
```diff
@@ -46,6 +46,12 @@ $sectionNavs = [
         <div class="value"><?php echo $escape($media->mediaType()); ?></div>
     </div>
     <?php endif; ?>
+    <?php if (is_numeric($resource->size())): ?>
+    <div class="meta-group">
+        <h4><?php echo $this->translate('Size'); ?></h4>
+        <div class="value"><?php echo sprintf('%s bytes', number_format($resource->size())); ?></div>
```
zerocrates: Possible issue with the use of `number_format` here... we probably should be using the locale-aware numberFormat, but we'd have to add that to our i18n helper to allow it work for people without intl... maybe not worth worrying about too much.

I don't think it should hold up merging but something to look at, possibly file an issue for.

application/view/omeka/admin/media/show-details.phtml
```diff
@@ -25,6 +25,13 @@ $item = $resource->item();
     </div>
     <?php endif; ?>
 
+    <?php if (is_numeric($resource->size())): ?>
+    <div class="meta-group">
+        <h4><?php echo $this->translate('Size'); ?></h4>
+        <div class="value"><?php echo sprintf('%s bytes', number_format($resource->size())); ?></div>
```
zerocrates: Needs to be translatable

application/view/omeka/admin/media/show-details.phtml
```diff
@@ -25,6 +25,13 @@ $item = $resource->item();
     </div>
     <?php endif; ?>
 
+    <?php if (is_numeric($resource->size())): ?>
```
zerocrates: The value must either be a number or null, right? Seems these should be "not null" checks... is_numeric more appropriate when the value is a number-ish string.

application/view/omeka/admin/media/show-details.phtml
zerocrates: This is more lines than i intended to grab: the `%s bytes` string specifically.

application/view/omeka/admin/media/show-details.phtml
jimsafley: Doctrine returns string for bigint data types.

zerocrates: ok then

PR #1253, opened 2018-05-06
Various small cleaning to cherry-pick

Daniel-KM:
Various small cleaning to cherry-pick.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ new, clean checkout.
 * `phpcli_path` Default is to attempt to detect correct path to PHP. Use this option to specify a path if needed in your server configuration. For example: 
 ```
     'cli' => array(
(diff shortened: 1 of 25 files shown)


zerocrates:
Hey Daniel, I took 10 of these as-is, and one more partially (I did not take the portions of "Fixed uses and docblocks" that added `use` lines just for use in docblock annotations.)

The readme I think we'll just drop that section as duplication of the manual, and I'm not sure I prefer the new structure from "Simplified some lines..." so I didn't take that one.

Lots of good stuff here, including a couple doozies from simple namespace stuff. Thanks!

Daniel-KM:
Ok, fine. So for  `use` used for docblock annotations, the classes should be fully referenced inside the blocks.

PR #1255, opened 2018-05-11
Use first available generated thumbnail instead of placeholder

mjlassila:
When an item has multiple media attached and some of the media have  generated thumbnails and some  of media are non-visual and are using the default thumbnails, it would be sensible default to  use the first available proper thumbnail instead of the default thumbnail of first available file. This patch makes Omeka to use the first available proper thumbnail if the mediatype is image/* or application/pdf.

### DIFF
--- a/application/src/Api/Representation/ItemRepresentation.php
+++ b/application/src/Api/Representation/ItemRepresentation.php
@@ -76,8 +76,16 @@ class ItemRepresentation extends AbstractResourceEntityRepresentation
     {
         // Return the first media if one exists.
         $media = $this->media();
-        return $media ? $media[0] : null;
-    }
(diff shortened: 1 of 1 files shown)


zerocrates:
Apologies for not interacting on this much much sooner.

I don't think having this depend on the MIME type is a good move here... better if anything to look at whether the media has generated thumbnails at all.

Given the changes in Omeka S since this was proposed, custom thumbnail upload/selection most of all, I'm not sure this is as relevant of an issue as it was then. The merge also doesn't apply cleanly after all this time, so I think the best move here is to close the PR.

If you (or other observers) are still looking to customize which is the "primary" media beyond listing it first, feel free to open a new issue or PR. I could see a "first with thumbnail" strategy or explicit selection of which media is "primary" as possible workable options.

PR #1260, opened 2018-05-22
Genericized batch actions js for browse pages.

Daniel-KM:
I have new modules to export selected resources (for example Zotero and ods), so it is simpler to manage when the batch actions in admin.js is simplified.


zerocrates:
`data-select="selected"` here just refers to options that should be disabled when there are no selections and enabled when there are? For something that's just on/off like that I'd go for a class rather than a data attribute.

We really also should be using the HTML5 "bare" form for `disabled` and `prop` rather than `attr` in the JS, but those are preexisting problems there.

Daniel-KM:
I set a class instead of a data.

Daniel-KM:
Rebased.

Daniel-KM:
Fixed disabled, not the js.

Daniel-KM:
Rebased on 1.3.

Daniel-KM:
Rebased on 1.3.

Daniel-KM:
Rebased on 1.4.

Daniel-KM:
Merged for laminas.

Issue #1252, opened 2018-05-03
Sql index for the columns value and type in table value


Daniel-KM:
Is there a reason why there is no sql indexes for the columns value, type or even uri in the table value?

zerocrates:
One of my own personal installation has had indexes on `value` and `uri` since far back when when we would have been working on the data model more, so I'm tempted to say it's basically just an oversight.

type, value, lang, and uri are all probably reasonable to index.

zerocrates:
Okay, so I _did_ add the value and uri indexes at some point... I'll have to get back to you on this one. There's clearly an inconsistency floating around somewhere.

zerocrates:
This is just my own misremembering: we _do_ have indexes on value and uri (that would be _quite_ the oversight to not have indexing at all here).

We have to do these indexes outside the normal Doctrine system because Doctrine doesn't provide us a way to do size-limited indexes on these fields, but size-limiting is required for the indexes to be legal. So, there's an event that's used to just manually specify the SQL for the value table: Omeka\Db\Event\Listener\CreateTableOverride.

Funny the things you can forget in 2 years.

zerocrates:
As a final note, if you don't _actually_ have indexes on `value` and `uri` in your value table, then that's an issue specific to your install (we do have them in the install sql and they were added in a migration back when they appeared as well).

Of course, we can look at indexing type (and possibly lang) as well.

Daniel-KM:
Thanks you very much for all your notes. Yes, most of my installs have it, not the one that I checked before posting this issue, neither the entity, but it is in the main schema.sql, so there is no issue except if you want to add some indexes.

PR #1262, opened 2018-05-28
Added column "data" to entity Value to simplify search by type.

### LINKED ISSUE #1241: Literal, resource and uri for values
Values can manage three types of values: literal, uri and resources (that are an exception to manage internal uri in fact). But the custom vocab and the value suggest modules use the `type` to set their own type: the id of the vocabulary that is used, of the original source of the suggestion. 

In reality, custom vocab values are literal values with a list of authorities, and suggested values are uri values with a specific service provider. So this additional metadata should be managed in a new column "data" or "source" of the table value (or in an external one-to-one table). 

In such a way (no mix of the type and the source), it will be possible to search or to display "literal" or "uri". Currently, a search on "literal" doesn't return custom vocabs values, and a search on "uri" doesn't return suggested values.

Daniel-KM:
Fixes #1241 and #1252 , so now it's possible to search `literal` only (and Custom Vocab values are literal) or `uri` only (and Value Suggest values are uri).


Daniel-KM:
See an issue related to the choice of the type on https://forum.omeka.org/t/module-solr-and-non-dc-fields/6500. And the first way (the one of the commit) is probably a better way to fix it.

Daniel-KM:
Full discussion is on https://github.com/omeka/omeka-s/issues/1241.

PR #1267, opened 2018-06-03
Don’t display SitePagePagination view helper for unique pages

### LINKED ISSUE #1266: Prev/Next text should not be displayed when the page is unique
See [“Page navigation” on the forum](https://forum.omeka.org/t/page-navigation/6186): when there is a unique page on a site, a end user does not understand what “Prev | Next” text means.

These “disabled links” should probably be hidden in this case.

pols12:
SitePagePagination view helper returns null if the page which calls it does not have any previous nor next page.

Fix #1266


pols12:
#1266 has been fixed with another way.

PR #1268, opened 2018-06-04
Added a test for importing jsonld via easyrdf (fail on php 7.2).

Daniel-KM:
See https://github.com/lanthaler/JsonLD/pull/92


zerocrates:
It's looking more and more like we'll have to at least temporarily fork to have a version of json-ld that fixes this... unfortunate but it is what it is, I guess.

zerocrates:
@patrickmj we should talk about this because as far as I can see you really should be getting an error because of this whenever you try to parse JSON-LD. Just loading the file should be a fatal error.

PR #1274, opened 2018-06-14
Advanced search "starts with" and "ends with"

Daniel-KM:
The keys can be changed. 
A key for "matches" can be added too.


Daniel-KM:
Rebased on 1.3. Available via module [Next](https://github.com/Daniel-KM/Omeka-S-module-next).

PR #1277, opened 2018-06-18
Fixed assignment of item sets to an item when a list contains duplicate item set ids.

Daniel-KM:


zerocrates:
So this is specifically just a problem with including the same item set (or media) ID twice in an API create/update request?

Daniel-KM:
Yes, the issue is not related to the ui, but to the api. When the media and item sets are appended via multiple events, it is not possible to check the input, so it's better to check them at the end, in the adapter. The issue may exist for other entities.

Daniel-KM:
Rebased on 1.3.

Daniel-KM:
Rebased on 1.4.

Daniel-KM:
Merged with laminas.

PR #1279, opened 2018-06-25
Fixed readme.

Daniel-KM:


zerocrates:
Sure, this is good. In general I'm thinking we should pare down the readme in favor of this kind of reference to the documentation/manual.

PR #1280, opened 2018-06-25
Feature/admin views

Daniel-KM:
The first commit is more a fix than a feature. Sometime in the admin board, the links goes to the public front-end, sometime, not. So the first commit add an icon to avoid issue and avoid to open in the same page.

The second fix is just for ergonomy: there is now a direct link to the public page, like in Omeka Classic.


Daniel-KM:
I rebased it on master, so the issue will be fixed. Note that I link may be kept, but it should be a link to the site admin site, not to the public site.

Daniel-KM:
The issue on Travis is strange, it works on php 7 and php 7.2, but not on php 7.1. Probably an internal error that is not related to the rebase.

zerocrates:
Yeah, I told Travis to do that 7.1 job again so it should clear itself up.

PR #1281, opened 2018-06-25
Allowed to return search in a random order like in Omeka Classic.

Daniel-KM:
For example, it allows to have a page block `Browse preview` or a carousel with a random list.


Daniel-KM:
Rebased on 1.3. Available via module [Next](https://github.com/Daniel-KM/Omeka-S-module-Next).

PR #1283, opened 2018-07-03
Feature/user bar admin links

Daniel-KM:
Make the user bar more useful and fix https://forum.omeka.org/t/add-a-link-to-modify-the-current-item/6580.
The css can be improved.


Daniel-KM:
Test fixed.

Daniel-KM:
Improved and rebased on 1.3. Available via module [Next](https://github.com/Daniel-KM/Omeka-S-module-Next).

zerocrates:
@Daniel-KM your change here to the default layout is just to make the user bar appear on the error pages? I was thinking the homepage possibly but it won't render if there's no site. It's not strictly related to the resource links that are the main thrust of the PR, in other words.

@kimisgold can you take a look at this? The CSS selector for the separator(s) definitely looks like a concern.

Daniel-KM:
On my point on view, the default layout would have the user bar, because there is a settings to disable/enable it. But it's at your choice, all my pull requests can be improved or cleaned. I can move the css file and the layout changes in another branch. 

And yes, I'm not good in css... And in practice, I have a last point that is "edit", that is fine, but sometime, I just need to show the item in admin page, so instead of edit, maybe replace with icons browse / show / edit / delete.

zerocrates:
We've taken the existing version...

As for other actions, I could see Show vs Edit... but icons means loading the icon font which I'd probably rather avoid just for this. Plus I think Delete is probably a bridge too far for things to include on the bar.

PR #1284, opened 2018-07-03
Moved scripts into footer for performance, accessibility, security, and semantic reasons.

Daniel-KM:
Same remarks than https://github.com/omeka-s-themes/default/pull/12.

The same can be done for admin, but I didn't checked all common modules yet.

### DIFF
--- a/application/view/layout/layout.phtml
+++ b/application/view/layout/layout.phtml
@@ -16,7 +16,6 @@ $this->trigger('view.layout');
         <?php echo $this->headTitle(); ?>
         <?php echo $this->headLink(); ?>
         <?php echo $this->headStyle(); ?>
-        <?php echo $this->headScript(); ?>
     </head>
     <?php echo $this->htmlElement('body')->appendAttribute('class', 'minimal'); ?>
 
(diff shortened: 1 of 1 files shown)


zerocrates:
Hmmm, I'm iffy on moving the headScript call as I'm pretty sure there are still scripts out there that are not tolerant to loading outside `head`. Plus, I know of the performance concerns but I believe they're largely eliminated by the `defer` attribute if used, if I'm not mistaken... of course we're not doing that at all currently either.

Perhaps my biggest issue would be just the "naming" issue: this is the headScript helper and it therefore implies (rather, explicitly states) that scripts queued on it will be rendered in the `head`. There is another Zend-provided helper named such that it specifies end-of-body tag placement, and I'd be fine with adding that to the layout(s), thereby giving us the choice of head or end-of-body.

The other parallel changes I've seen you propose to the themes as part of this, like breaking out redundant inline JS into static assets, I'm unambiguously in favor of.

Daniel-KM:
If a js cannot be loaded either in head or in body, it should be fixed. In the themes I use, I put the headscript in the body and it works fine (I fixed my modules too).
The aim is to have all the scripts at the same place and to remove inline ones. If some modules use the head script and some other ones the body script, some js will be duplicated and it won't be cleaner. So it's better to choose either the head or the body.

Daniel-KM:
Outdated with #1479.

PR #1291, opened 2018-08-13
Various small fixes that can be cherry picked

Daniel-KM:


application/src/Api/Adapter/AdapterInterface.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Omeka\Api\Adapter;
 
+use Omeka\Api\Representation\RepresentationInterface;
```
zerocrates: This is just for its appearance in the docblock for getRepresentation, yeah? Would prefer to fix these by fully-qualifying within the docblock rather than a `use`.

application/src/Controller/ApiController.php
```diff
@@ -243,8 +242,8 @@ protected function getErrorResult(MvcEvent $event, \Exception $error)
      *
      * @param string
      * @return mixed
-     * @throws Exception\DomainException if no JSON decoding functionality is
-     *     available.
+     * @throws Exception\InvalidJsonException if no JSON decoding functionality
```
zerocrates: The note about this specific exception actually probably dates to when we used Zend's JSON component here. We should update this `@throws` but need to fix the actual comment as well, since it actually throws on JSON decoding errors or if the content is a scalar.

application/Module.php
```diff
@@ -25,9 +25,6 @@ class Module extends AbstractModule
      */
     const OMEKA_VOCABULARY_TERM = 'o';
 
-    /**
-     * {@inheritDoc}
-     */
```
zerocrates: I'm fine with removing these, I think...

This pattern originally came from following Zend Framework style guidance but they really are useless at best.

application/src/Mvc/Controller/Plugin/UserIsAllowed.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace Omeka\Mvc\Controller\Plugin;
 
+use Omeka\Api\ResourceInterface;
```
zerocrates: Same objection to the `use`s in this commit.

application/src/Mvc/MvcListeners.php
```diff
@@ -144,7 +144,7 @@ public function bootstrapLocale(MvcEvent $event)
      * Redirect all requests to install route if Omeka is not installed.
      *
      * @param MvcEvent $event
-     * @return Zend\Http\PhpEnvironment\Response
+     * @return \Zend\Stdlib\ResponseInterface
```
zerocrates: The responses being used here are actually the HTTP type indicated, and there's quite a bit of their functionality only available in those more specific subclasses, so we probably don't want to genericize this down to just the ResponseInterface.

application/src/View/Helper/AssetUrl.php
```diff
@@ -83,7 +83,7 @@ public function __invoke($file, $module = null, $override = false, $versioned =
             && ($module === 'Omeka' || array_key_exists($module, $this->activeModules))
         ) {
             $themeId = $this->currentTheme->getId();
-            $filepath = sprintf(self::THEME_ASSETS_PATH, OMEKA_PATH, $themeId, $file, '');
+            $filepath = sprintf(substr(self::THEME_ASSETS_PATH, 0, -5), OMEKA_PATH, $themeId, $file);
```
zerocrates: This isn't actually broken in its current form, is it? And doesn't the `-5` substring chop off too much of the template here, losing the actual filename part?

I don't like the clunkiness of the current sprintf here either, though... it'd probably be better to drop the extra `%s` from all the templates and just factor out the addition of the version GET parameter to another function or something like that.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -277,7 +277,7 @@ public function values()
      *   by default.
      * - lang: (null) Get values of this language only. Returns values of all
      *   languages by default.
-     * @return RepresentationInterface|mixed
+     * @return ValueRepresentation[]|mixed
```
zerocrates: I guess this should really be `ValueRepresentation|ValueRepresentation[]|mixed`. Maybe we should just give up and specify it as `mixed`. If we're going to specify though, a lone `ValueRepresentation` is the "usual" case so should probably be first.

application/view/omeka/admin/item-set/show.phtml
```diff
@@ -13,30 +13,30 @@ $viewItemsLink = $this->url(
 if ($owner = $itemSet->owner()) {
     $ownerText = $escape($itemSet->owner()->name());
 } else {
-    $ownerText = $this->translate('[no owner]');
+    $ownerText = $translate('[no owner]');
```
zerocrates: Am I just missing where `$translate` is defined in most of these? Is it just "leaking" in from some previous view?

application/src/Api/Adapter/ApiResourceAdapter.php
```diff
@@ -17,7 +17,7 @@ public function getResourceName()
 
     public function getRepresentationClass()
     {
-        return 'Omeka\Api\Representation\ApiResourceRepresentation';
+        return \Omeka\Api\Representation\ApiResourceRepresentation::class;
```
zerocrates: Official guidance on class names as strings is to omit the leading backslash since class names in strings are always presumed to be fully-qualified.

If we have places where we _are_ including the leading backslash in a string, I'd consider _that_ something we should fix, but not in this direction.

zerocrates: Wait, I misread these I guess. Of course using `::class` is fine. I'm sure I'd seen them as just backslashes in the string but I must be imagining things.

This is fine.

application/src/Api/Adapter/AdapterInterface.php
Daniel-KM: Ok. Fixed in next commit.

application/src/Controller/ApiController.php
Daniel-KM: Ok. Fixed in next commit.

application/Module.php
Daniel-KM: Yes, the `@inheritDoc` should not be used alone, according to the phpdoc doc (https://docs.phpdoc.org/guides/inheritance.html), but some applications set it by default.

application/src/Mvc/Controller/Plugin/UserIsAllowed.php
Daniel-KM: Ok. Fixed in next commit.

application/src/Mvc/MvcListeners.php
Daniel-KM: Ok. Fixed in next commit.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
Daniel-KM: When there are multiple possible outputs, it's better to use the most frequent one, because this is the one that is used by the editor, and `mixed` is not helpy in that case, so I prefer to use the three possible values here. Fixed in next commit.

application/view/omeka/admin/item-set/show.phtml
Daniel-KM: This is a small optimization, `translate` is used very often, so it should be used and set in most of the theme file. I check if it is missed in one file. Fixed in next commit.

application/src/Api/Adapter/ApiResourceAdapter.php
Daniel-KM: Use a class is simpler to use in the editor.

application/config/module.config.php
```diff
@@ -434,32 +434,32 @@
...
+            Form\Element\ResourceSelect::class => Service\Form\Element\ResourceSelectFactory::class,
+            Form\Element\ResourceClassSelect::class => Service\Form\Element\ResourceClassSelectFactory::class,
+            Form\Element\PropertySelect::class => Service\Form\Element\PropertySelectFactory::class,
+            Form\Element\ItemSetSelect::class => Service\Form\Element\ItemSetSelectFactory::class,
+            Form\Element\SiteSelect::class => Service\Form\Element\SiteSelectFactory::class,
+            Form\Element\LocaleSelect::class => Service\Form\Element\LocaleSelectFactory::class,
+            Form\Element\Recaptcha::class => Service\Form\Element\RecaptchaFactory::class,
+            Form\Element\HtmlTextarea::class => Service\Form\Element\HtmlTextareaFactory::class,
+            Form\Element\Ckeditor::class => Service\Form\Element\CkeditorFactory::class,
+            Form\Element\CkeditorInline::class => Service\Form\Element\CkeditorInlineFactory::class,
```
zerocrates: Oh, and I didn't notice this my first time through either... I don't know about these that switch the _keys_ to using the `::class` construction... even though these are equivalent to their class names here, I'm not sure we want to be doing that. `::class` on the _values_, since they actually are used as class names by the system, are OK/desirable (as in the other changes to this file), but I don't think we want to do this key-side change, at least not without thinking it over some more.

zerocrates:
Other than the last comment I just made, this all looks good to me.

application/config/module.config.php
Daniel-KM: Ok, I reverted it in last commit.

zerocrates:
OK. I rebased/squashed your fix commits into the originals and dropped out the one total reversion, but this is all merged.

PR #1292, opened 2018-08-15
Some more various small fixes that can be cherry picked...

Daniel-KM:


application/src/Api/Adapter/ItemSetAdapter.php
```diff
@@ -61,12 +61,6 @@ public function buildQuery(QueryBuilder $qb, array $query)
         }
 
         if (!empty($query['site_id'])) {
-            $siteAdapter = $this->getAdapter('sites');
-            try {
-                $site = $siteAdapter->findEntity($query['site_id']);
-            } catch (Exception\NotFoundException $e) {
-                $site = null;
-            }
```
zerocrates: Though the site isn't used here, this is intended to ensure that you cannot perform a query against a private site you don't have access to.

application/src/Api/Adapter/ItemSetAdapter.php
```diff
@@ -2,7 +2,6 @@
 namespace Omeka\Api\Adapter;
 
 use Doctrine\ORM\QueryBuilder;
-use Omeka\Api\Exception;
```
zerocrates: This is actually used/required, I believe.

application/src/Api/Adapter/ItemSetAdapter.php
Daniel-KM: So it is not clear how it works, so a comment should be added at least.

zerocrates:
Merged.

PR #1304, opened 2018-09-14
Fix bug with "Filter by property" on item page.

kloor:
In Omeka v1.2.0, the dropdown box "Filter by property" on item pages presents an error if you select a specific property. I think that it is because the query statement was not being built correctly with the property, and this change works for me as a fix.


zerocrates:
Yeah, sure looks like a regression. An error I made when fixing a different bug, I think.

PR #1307, opened 2018-09-21
Private values

jimsafley:


jimsafley:
Testing here: #1311

jimsafley:
The addition of the visibility toggle pushes the `.input-body` div down, causing a white bar to appear under the textarea:

value

Removing `background-color: #fff;` from the div removes the bar, but it breaks the style for resource and URI value types:

values

PR #1319, opened 2018-10-11
Fixed default height of textarea.

Daniel-KM:
Often, the text areas have more than 5 lines, but we cannot resize them in the browser.
The same occurs for the inline ckeditor (not fixed here).


zerocrates:
Hmm, I haven't checked this out, but I assume it has the effect of also increasing the default sizes of all the textareas? We probably don't want to do that.

I don't think we'd have any issue with a significantly expanded max-height, though. Unconstrained is an option as well... I'm not as sure about it.

Daniel-KM:
There are two default sizes, the number of lines of text and the size of the text area itself. The default number of lines is still 5 lines, so the display doesn't change, but I can now display all rows if needed, because it is not limited by the max height.

zerocrates:
Usually our defaults are for 2 lines, not 5, though.

Daniel-KM:
Anyway, 120px is too small for a text area and I don't understand why there is a limit, event if it can be small when the page is loaded.

zerocrates:
Kim's currently on leave so you're at the mercy of my decisionmaking at the moment on this one...

I'd be more than happy to make (or accept) a change that increases the current maximum but I'm hesitant to remove it completely, and definitely don't want to use `height` and thereby change the default sizes that are currently there.

zerocrates:
We've gone ahead and decided to remove the upper height limit completely.

PR #1331, opened 2018-10-30
Fix/translate plugin

Daniel-KM:
I removed these commits from the fix/ongoing one, because many files are modified, but it's just a replace. 

The use of $translate() is a little quicker to use than $this->translate(), but the main advantage is for the theme designers, because it is simpler to write.

### DIFF
--- a/application/view/common/advanced-search/item-sets.phtml
+++ b/application/view/common/advanced-search/item-sets.phtml
@@ -1,4 +1,5 @@
 <?php
+$translate = $this->plugin('translate');
 // Prepare the item set queries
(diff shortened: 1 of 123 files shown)


zerocrates:
OK. I knew I hadn't pulled those in so I was momentarily confused where they'd gone in your previous rebase.

Daniel-KM:
And I didn't check if the strings that use the function `$translate()` can be extracted or not.

Daniel-KM:
Rebased.

Daniel-KM:
Rebased on 1.3.

PR #1334, opened 2018-11-05
Updated install schema.

Daniel-KM:


zerocrates:
We fixed this slightly differently (this change would have been overwritten by future schema updates), but thanks for flagging it.

PR #1335, opened 2018-11-05
Upgrade/easyrdf

Daniel-KM:
[EasyRdf](https://github.com/njh/easyrdf) is still in version 9.0 since four years, but in fact, the version 0.10 alpha works a lot better. In particular, you can now import ontologies of any size (many MB) in some seconds only. So this simple upgrade fixes many issues for heavy or complex ontologies. 

The second commit is a pure fix: i want to import properties that are annotation property too, but this is not possible in this case (for example in the skos ontology, `skos:note` is a `owl:AnnotationProperty , rdf:Property`, so it should be imported, but it can't because only the first type was checked). 

On this subject, I like to add a checkbox to let the user to import all properties, included annotation properties (that I need for one project).

### DIFF
--- a/application/src/Stdlib/RdfImporter.php
+++ b/application/src/Stdlib/RdfImporter.php
@@ -84,7 +84,7 @@ class RdfImporter
         }
 
         // Get the full RDF graph from EasyRdf.
-        $graph = new \EasyRdf\Graph;
+        $graph = new EasyRdf\Graph;
         switch ($strategy) {
             case 'file':
                 // Import from a file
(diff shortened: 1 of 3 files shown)


Daniel-KM:
Rebased on 1.4.

Daniel-KM:
The part on the types should probably be skipped.

Daniel-KM:
Fixed in https://github.com/omeka/omeka-s/commit/8d1476bcc8f2ec51126a44ea7497025ea1dbcb3b

Issue #1293, opened 2018-08-24
Site pages and languages


Daniel-KM:
About multilanguage, I need to add a language to pages, so maybe a new field should be added in table site_page or site_page_block, but I don't know yet which one is better. Do you have plan about this point and about the global pages (like the about page)?

This issue may be related to the multilanguage site issue (https://github.com/omeka/omeka-s/issues/898), and there is no clear way to manage such a site with Omeka S, and there may be multiple cases (the two main are one site by language and multiple languages by site).

Daniel-KM:
On this important question, I wonder if a language setting can be attached to two things: 
- the site : a simple site setting to set the default language of a site, for anonymous users, so it will be possible to resolve many questions of the forum (https://forum.omeka.org/t/bilingual-content-internationalisation/2116, https://forum.omeka.org/t/internationalized-omeka-s/3154/22, etc. and here (#898), and in, particular to make each site multilingual, not only a site by language.
- the page: a new column in site_page, so the pages will have a language, so it will be possible to make a multilanguage module like in Omeka Classic, either at site level in the case of a multilanguage site, or in the whole in the case of a site by language).

I can push a request on  that, but I need a confirmation that it is a good structural choice.

zerocrates:
We did some discussion of this topic in the the attached PR, closing by reference to that.

PR #1347, opened 2018-11-20
Feature/locale page

### LINKED ISSUE #898: Figure out multilanguage sites
We have a way to tag language for values, but don't yet have a system for using them, other than the signal about what language a value has set. Seems like some way to 1) set a language, or languages to display and 2) filter based on that setting might be in order. Alternatively, people have suggested a way to have multiple sites that use different languages (which sounds like a site setting). Either of those approaches sound like having the theme be able to know what language(s) to display and filter values accordingly.

One feature we know people want is on-the-fly language switching, without being logged in and without changing browser setting. However, a system that _can_ work with browser settings also makes sense for the filtering.

See convo on the forums https://forum.omeka.org/t/internationalized-omeka-s/3154

Daniel-KM:
This feature is the required first step to make Omeka full multilingual, as many users want. 

There are two main ways to manage localization: one site by language, or multilanguage by site. The first way has many limitation, in particular it implies to have two urls for the same pages and documents. The second way is the normal way we found on many cms. 

Because there is a new value in the database for the site pages, it can be used by future modules easily, unlike in Omeka Classic.

Fix #898, #1293, https://forum.omeka.org/t/bilingual-content-internationalisation/2116, https://forum.omeka.org/t/internationalized-omeka-s/3154/22, etc.

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -13,7 +13,7 @@ class Module extends AbstractModule
     /**
      * This Omeka version.
      */
-    const VERSION = '1.2.1-alpha4';
+    const VERSION = '1.2.1-alpha.5';
 
     /**
      * The vocabulary IRI used to define Omeka application data.
--- a/application/Module.php
+++ b/application/Module.php
@@ -13,7 +13,7 @@ class Module extends AbstractModule
     /**
      * This Omeka version.
      */
-    const VERSION = '1.2.1-alpha.5';
+    const VERSION = '1.3.0';
 
     /**
      * The vocabulary IRI used to define Omeka application data.
--- a/application/asset/js/admin.js
+++ b/application/asset/js/admin.js
@@ -190,7 +190,7 @@ var Omeka = {
     },
 
     // @see http://stackoverflow.com/questions/7035825/regular-expression-for-a-language-tag-as-defined-by-bcp47
-    // Removes `|[A-Za-z]{4}|[A-Za-z]{5,8}` from the "language" portion becuase,
+    // Removes `|[A-Za-z]{4}|[A-Za-z]{5,8}` from the "language" portion because,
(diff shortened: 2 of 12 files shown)


Daniel-KM:
For  #1349 (1.3) ?

zerocrates:
It won't go into 1.3.

I'm not sure about this in general as the solution though... you note the differences between the basic strategies for doing multilingual within the sites, but I'm not sure language tagging at the page level gets us all that far... you'd still have basic problems with things like the slug of the site, the title of the site (also the slug of the pages which would still have to vary between different language "versions" of the same page, the new "summary" of the site, the strings in the navigation, the actual resource browse/show views, and the theme settings (things like text).

Obviously you've framed this as a first step so it's not exactly a problem that it doesn't deal with those things but my concern is it's a first step down a road of adding significant complexity to all those areas (at least). The main advantage of the "site" level of language tagging that we currently have is that all those pieces of simple data within the site are covered: you can still only have one slug, one title, and so on, but since the site can declare its language that's fine. Of course this implies duplicating the site to have versions of it in different languages, but you're fundamentally doing that duplication no matter the model.

In my experience different URLs for different languages of the same content is pretty standard, but I'd be happy to hear about counterexamples or alternatives.

In short: I'm not against language tagging at the page level like you're doing here, but I'm just not sure how it fits into an eventual "complete" multilingual solution for S sites.

Daniel-KM:
Ok if you prefer the solution of one language by site, it's good to know, I'll fix my pr in that way. 
So this solution implies at least the possibility to have a "/" in the slug name in order to do `/my-site/fr` or `/my-site/en`, and/or in the page slug. This point refers too to the more general issue of web site management (use a domain by site, to have a main site).

PR #1354, opened 2018-12-07
Allow arbitrary keys to be passed into batch update job

jimsafley:
This brings "batch update all" in line with "batch update selected," which allows arbitrary keys to be passed in the request data. Otherwise, only the `replace`, `remove`, and `append` keys (those that declare the "collection action" request option) can be passed to and processed by the batch update job. This is a safe change because the data is still passed through `preprocessBatchUpdate()` in the adapter.


jimsafley:
Looking closer, it seems to me that we shouldn't allow arbitrary keys in the request data, and instead just pass arbitrary data through one of the three collection actions. That actually seems safer and more predictable.

PR #1357, opened 2018-12-22
Display linked resource Title instead of URI in Browse and Browse Preview

ManOnDaMoon:
Hello,

I'm suggesting a slight modification of how Browse and BrowsePreview templates display linked resources' heading and body.

Currently if the selected _body_ property is of type 'resource', Omeka-S displays its (non clickable) URI which is not very intelligible.
Also, when displaying a _heading_ property of type 'resource', Omeka-S displays the URI property actually linking to a different URL, which is not very web-standard friendly.

These modifications test for the property type, and if it is a linked resource, item, item set or media, it tries to retrieve the 'dcterm:title', else defaults to [Untitled].

If accepted, I guess these modifications should also be rolled out to themes that modify these templates.
Thanks in advance and happy holidays.

Laurent Thomas

### DIFF
--- a/application/view/common/block-layout/browse-preview.phtml
+++ b/application/view/common/block-layout/browse-preview.phtml
@@ -17,7 +17,9 @@ foreach ($this->resources as $resource):
         'resource:item',
         'resource:itemset',
         'resource:media'
-    ])) ? $headingTermValue->valueResource()->displayTitle() . " (" . $resource->value('dcterms:title') . ")" : $headingTermValue;
+    ])) ? $headingTermValue->valueResource()->displayTitle() . " (" . $resource->value('dcterms:title', [
+        'default' => $this->translate('[Untitled]')
+    ]) . ")" : $headingTermValue;
 
 
     $bodyTermValue = $resource->value($bodyTerm);
--- a/application/view/common/block-layout/browse-preview.phtml
+++ b/application/view/common/block-layout/browse-preview.phtml
@@ -12,12 +12,13 @@ foreach ($this->resources as $resource):
     $headingTermValue = $resource->value($headingTerm, [
         'default' => $this->translate('[Untitled]')
     ]);
-    $heading = ($headingTermValue && !is_string($headingTermValue) && in_array($headingTermValue->type(), [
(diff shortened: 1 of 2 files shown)


zerocrates:
This probably makes sense... it can be a little unclear what people _want_ to have happen when they've set a linkage as the "title" of a resource... I could foresee some confusion if, for example, the "original" item and the one linking to it appear in the same browse.

Regardless it's pretty likely that showing the URI isn't what's expected.

ManOnDaMoon:
An addition in order to lift confusion could be a visual indicator that the displayed title is a linkage, something similar to the 🔗 Unicode character, or the "external link" icon used throughout the admin panel.

ManOnDaMoon:
One simpler way to keep this from getting confusing would be to combine both linked resource and actual item title. E.g.: keep the listed item title in parenthesis, right behind the linked resource term value.

zerocrates:
We're looking into some changes around the title/summary selection that could affect this... but they're not ready to go at the moment.

One question: do you have a use case that actually uses "titles" like these? (i.e., the Title of resource A is actually a link to Resource B and that's what you _want_ it to be?) I'm having trouble figuring a situation where that's actually desired... it seems more or less doomed to be confusing.

Here's why I ask: part of what we're thinking about is separately storing the title for resources, so it's easier for us to get at a title for things like searching, or rendering links... but there's some question as to what kind of value should "count" as a title, and we've thought about simply excluding internal links and URIs from being considered the title at all (falling down to look for later literal values or simply showing "Untitled" if only links are provided in the title property).

ManOnDaMoon:
To be honest I don't know of any reasonable use case of my own for this. The only use I could think of was to use linked resources to "sync" titles among resources, i.e. have a unique resource holding a _literal_ title and using it as a _linked resource_ title for another set of resources.

ManOnDaMoon:
Hello, Is this is still admissible, I'll submit an update soon in order to fix conflicts with the recent updates. Should I rollback the change on headings though?

ManOnDaMoon:
This is now outdated

PR #1360, opened 2019-01-11
Displaying the number of media linked to the item

symac:
Has been asked by some colleagues and I believe it makes sense to have a quick overview of it.


zerocrates:
A couple quibbles:

- I'd prefer `count` over `sizeof`, the latter is much more rarely used.
- We're guilty of lazily not doing this in the existing code, but as this adds yet another call to the method, the array of media should be retrieved into a variable and all the other references to it (the if, this count, the foreach) should use the variable.

symac:
I had an issue with my branch and was not able to apply the requested changes to it so have created a new PR that takes into account the requested changes : #1391

PR #1362, opened 2019-02-06
Update sidebar-select.phtml

jaredhowland:
Allow any image type to be uploaded from the admin side of Omeka S (for example, `.svg` logos for themes).


jaredhowland:
More work needs to be done on this pull request before this will work because of a change needed in validating media types. I will work on that and recreate a pull request.

jaredhowland:
Found the validation error and fixed it to add `.svg` files. I'm closing this pull request with the complete solution.

PR #1363, opened 2019-02-06
Add `.svg` support for themes in Admin

jaredhowland:
I'd like to add support for uploading `.svg` and `.svgz` files on the admin side for themes (for logos and such). This pull request allows those files to be uploaded and updates the validator to allow them to work.


zerocrates:
I was a little leery about the previous version that extended to all image types but I think this one should be fine.

jaredhowland:
Awesome! This will make a theme I am working on much better and I think will open up better design opportunities for others who use `.svg` files in their designs.

PR #1370, opened 2019-02-19
Check for resource values with default type set. (Fix #1369)

### LINKED ISSUE #1369: Icon for using Omeka Item as value missing on admin (minor)
itemindicator

See Participant property.

Appears to be after 1.3 upgrade.

kimisgold:
When displaying a resource value, this checks to see if the value type contains the string 'resource' to catch values that have a default type set to resources. (These values output "resource:[resource name]".)


zerocrates:
Guess I don't need to merge this because you seem to have rebased/cherry-picked it in with your latest commit

PR #1372, opened 2019-02-21
Added a way to use psr-3 messages.

### LINKED ISSUE #1183: Plan to make log psr-3 compliant
In order to get translatable logs, and manageable logs with third party services, they must be psr-3 compliant (see details on http://www.php-fig.org/psr/psr-3/ or on my module ([Daniel-KM/Omeka-S-module-Log](https://github.com/Daniel-KM/Omeka-S-module-Log)). Is there a plan for that (it implies to modify all log messages that use sprintf)? Or will you accept a pull request for that?

Daniel-KM:
Allows to use psr-3 compliant messages : https://www.php-fig.org/psr/psr-3

When accepted, another pr will allow to use psr strings in [logs](https://github.com/Daniel-KM/Omeka-S-module-Log/blob/master/src/Formatter/PsrLogSimple.php).

Fixes #1183

### DIFF
--- a/application/src/I18n/Translator.php
+++ b/application/src/I18n/Translator.php
@@ -1,8 +1,9 @@
 <?php
 namespace Omeka\I18n;
 
-use Omeka\Stdlib\Message;
+use Laminas\I18n\Translator\TranslatorAwareInterface;
 use Laminas\I18n\Translator\TranslatorInterface;
+use Omeka\Stdlib\MessageInterface;
 
 class Translator implements TranslatorInterface
 {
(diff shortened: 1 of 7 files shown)


zerocrates:
What kinds of situations are you envisioning where you might use this new message class outside the messenger or the logger? I'd really prefer to see it use the same code path as is currently used in the messenger for translating the sprintf-based messages... in other words I'd be happier with something that altered the existing Message and created an interface the two message types would share (something like `getMessage()` and `interpolate()` or whatever it would be called), and altered our translator to check for that interface rather than Message.

Daniel-KM:
The main usage is messenger and logger. I don't have any other usage currently.

Daniel-KM:
In fact, there is another usage: to write e-mails (config form of module GuestUser), where string placeholders are simpler to use by end users.

Daniel-KM:
Rebased on 1.4.

Daniel-KM:
Rebased on 2.0.

Daniel-KM:
Merged for laminas.

zerocrates:
OK, returning to this after many years.

I've pushed a branch with a version of this that I'd be happier with: psr_message_2. It basically turns the message interface, and both message concrete classes, into black boxes as far as the rest of the system is concerned. You can call toString on them, or pass a translator and call translate, and each message handles whatever interpolation that requires. This also means removing the translation-aware and auto-translating features of the PSR message class, with the idea being that you just either pass the message to a translator, or don't, for both types of message. Along with this, I shrink the interface down to just include those key methods.

Let me know if this minimized version suits your purposes still.

zerocrates:
The shared MessageInterface that both messages implement already extends from JsonSerilaizable, so there's no need to have an `implements JsonSerializable` in the message classes themselves.

I'm going to merge (and probably squash?) the refactored branch from earlier. My inclination is that the method aliases and name changes and so on aren't really necessary but I'm open to being persuaded on that, I guess.

zerocrates:
That's merged. I think I'll close this PR but you're of course welcome to bring over commits into another one, or just suggest that I grab them.

Obviously the piece we don't have here is the logger itself. Laminas provides a PSR-compatible interface for the logger so my current plan there is just to have a second service `Omeka\PsrLogger` which uses that compatible interface but backs onto the same logging backend.

In the future if/when we had to or wanted to move to monolog, we could do the same in reverse, provide a simple facade to the Zend/Laminas logger interface that just sends to the "real" PSR logger.

PR #1375, opened 2019-02-28
Add configurable password requirements via config file

jimsafley:


zerocrates:
wow this does... so much more than i was envisioning bothering with

zerocrates:
I was going to complain about un-guarded use of mbstring but we're already doing that in at least one place elsewhere... we should probably add mbstring to the extensions we check for in the installer.

jimsafley:
Without `mb_strtoupper()` and `mb_strtolower()` multibyte characters that have diacritics (e.g. Ü, Ã) do not convert to upper or lowercase. This throws off the string comparisons that are necessary to count upper and lowercase characters in the string.

jimsafley:
@kimisgold The [password-config](https://github.com/omeka/omeka-s/tree/password-config) branch needs attention, particularly on the output of `Omeka\View\Helper\PasswordRequirements` which is rendered on, among other places, the user edit form. You can adjust the output by editing the `password` array in application/config/module.config.php.

application/view/omeka/admin/user/edit.phtml
```diff
@@ -27,6 +27,7 @@ $formElement = $this->plugin('formElement');
 </fieldset>
 <fieldset id="change-password" class="section">
     <?php echo $this->formCollection($form->get('change-password'), false); ?>
+    <?php echo $this->passwordRequirements(); ?>
```
kimisgold: In the default list of password requirements, is there currently a scenario where we have more than the one requirement? Like in this view:

Screen Shot 2019-03-22 at 2 45 09 PM

If we only ever have the one requirement, I'm not sure if it should be a list.

jimsafley: There can be up to 5 requirements listed there.

kimisgold: How are those requirements triggered to appear?

jimsafley: Whenever `min_length`,  `min_lowercase`, `min_uppercase`, `min_number`, or `min_symbol` are set to something other than `null`.
```  
  'password' => [
        'min_length' => 6,
        'min_lowercase' => null,
        'min_uppercase' => null,
        'min_number' => null,
        'min_symbol' => null,
        'symbol_list' => '`~!@#$%^&*()-=_+[]\{}|;:",./<>?\'',
    ],
```

PR #1385, opened 2019-03-10
Add default visibility and apply it to new content

ManOnDaMoon:
Hello,

First, there are currently two different behaviors throughout Omeka S when it comes to default visibility:
* the visibility setting is by default set to 'Public' on new items and new item sets forms
* the site creation form has the visibility setting set to 'Private' by default

Secondly, in use cases where users wish to keep all content private (e.g. to prevent anonymous API access), they need to toggle the visibility setting each time instead of having a global option to do this. The nature of the form input (an icon a bit displaced above the form) also helps forgetting to apply the setting once in a while (more often than not, actually).

This PR addresses the two behaviors by adding a global setting to have a default visibility option that consistently applies to content and sites. Please note that the visibility can still be changed in each form, this is only about what default value is selected.

Please do not hesitate to ask if this requires any adjustments.

Have a nice day

Laurent

### DIFF
--- a/application/src/Form/SettingForm.php
+++ b/application/src/Form/SettingForm.php
@@ -243,7 +243,7 @@ class SettingForm extends Form
             'type' => 'Checkbox',
             'options' => [
               'label' => 'Default content visibility to Private', // @translate
-              'info' => 'If checked, all items and item sets created will have their visibility set to private by default.' // @translate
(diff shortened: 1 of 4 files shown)


zerocrates:
This is a good feature but I don't think we'll have it in the immediately-oncoming release... I'm not sure exactly how (if at all) we want this to interact with API defaults and so on.

PR #1387, opened 2019-03-13
Add default homepage site setting

### LINKED ISSUE #1286: Select site's default homepage without adding to navigation
Currently, a site's default homepage is determined by the first item within the site's navigation. It's a frequent pattern to have the site title link to the homepage, separate from the rest of the navigation. I'd like to have a site setting where I can select a default homepage that would be linked by the site title, rather than having to have an explicit "Home" navigation item.

jimsafley:
Fixes #1286


zerocrates:
@kimisgold can you take a look at this?

My only real question was whether this control should be in Settings, as here, or in Navigation (a la Classic).

kimisgold:
For the moment, I think it makes the most sense within Navigation. It would be easy to find there, as opposed to getting buried as a field within general site settings or having to create a navigation subsection within Settings with only one field.

application/src/Entity/Site.php
```diff
@@ -41,6 +41,12 @@ class Site extends AbstractEntity
      */
     protected $navigation;
 
+    /**
+     * @ManyToOne(targetEntity="SitePage")
```
zerocrates: Should be `OneToOne`

application/src/Api/Adapter/SiteAdapter.php
```diff
@@ -89,6 +89,15 @@ public function hydrate(Request $request, EntityInterface $entity,
...
         }
+        if ($this->shouldHydrate($request, 'o:homepage')) {
+            $homepage = $request->getValue('o:homepage');
+            if (isset($homepage['o:id']) && is_numeric($homepage['o:id'])) {
+                $homepage = $this->getAdapter('site_pages')->findEntity($homepage['o:id']);
+            } else {
+                $homepage = null;
+            }
+            $entity->setHomepage($homepage);
+        }
```
zerocrates: Should have a sanity check here or in validate that the homepage is actually one of the site's pages.

PR #1391, opened 2019-03-22
Display the number of medias attached to the item.

symac:
This replaces the #1360 where some fixes had been asked by @zerocrates but I had deleted my branch meanwhile and am not good enough with git to update it then, so have just created this new PR.


zerocrates:
Ah, great, thanks for making those changes.

PR #1396, opened 2019-04-10
Quick fixes for translations

ManOnDaMoon:
Hello,

A quick PR for fixing 2 missing translation tags:
- in site settings page, the "Restrict to templates" label was not marked for translation
- in search filters view helper, the resource class label was not translated

Thank you in advance and have a nice day

Laurent


ManOnDaMoon:
A quick explanation about a07b693a60924fdbb7a023d3fd3d80a6e559e96f: the distinction between _login_ and _log in_ is a bit stronger in some languages than in english. Especially on the login page, having 'Login' as the title is like displaying 'A login' instead of requesting to sign in to the application. The verb form will be less ambiguous here.

PR #1400, opened 2019-04-15
Made entity fluid.

Daniel-KM:
Made entity fluid. Another commit will be added to upgrade all places where entities are used when committed.


zerocrates:
Hmm... I'm not sure if we want to use a fluent interface here... I don't think it's terribly common for Doctrine entities, and they're by design dead-simple boilerplate.

zerocrates:
I think I'm still feeling the same as my earlier comment: this isn't something I think we're really interested in changing.

PR #1405, opened 2019-04-30
Use faster counts for items and item sets on representations.

kloor:
Related to #1404, we're working on improving admin interface responsiveness when having hundreds of thousands of items. It appears that one of the things causing item counts to be retrieved slowly is the amount of data retrieved and steps taken with that data after it is retrieved and before it is counted.

One of the easier changes found to remedy this was to specify that counts should be taken from a single field via the returnScalar option, which also prevents the finalize tasks from being performed. This pull request changes Item Sets, Properties, Resource Classes, Resource Templates and Users to retrieve item counts with that option. Users are also changed to retrieve item set counts the same way.

With this change, item counts remained the same but were performed noticeably faster. These methods appear to only be used for retrieving counts in the admin interface. I'm not certain if finalize would ever impact the counts.

### DIFF
--- a/application/src/Api/Representation/ItemSetRepresentation.php
+++ b/application/src/Api/Representation/ItemSetRepresentation.php
@@ -39,9 +39,8 @@ class ItemSetRepresentation extends AbstractResourceEntityRepresentation
     {
         $response = $this->getServiceLocator()->get('Omeka\ApiManager')
             ->search('items', [
-                'item_set_id' => $this->id(),
-                'limit' => 0,
-            ]);
+                'item_set_id' => $this->id()
+            ], ['returnScalar' => 'id']);
         return $response->getTotalResults();
     }
 
--- a/application/src/Api/Representation/ItemSetRepresentation.php
+++ b/application/src/Api/Representation/ItemSetRepresentation.php
@@ -39,7 +39,7 @@ class ItemSetRepresentation extends AbstractResourceEntityRepresentation
     {
         $response = $this->getServiceLocator()->get('Omeka\ApiManager')
             ->search('items', [
-                'item_set_id' => $this->id()
+                'item_set_id' => $this->id(),
             ], ['returnScalar' => 'id']);
         return $response->getTotalResults();
     }
--- a/application/src/Api/Representation/PropertyRepresentation.php
+++ b/application/src/Api/Representation/PropertyRepresentation.php
@@ -28,8 +28,7 @@ class PropertyRepresentation extends AbstractVocabularyMemberRepresentation
                         'type' => 'ex',
                     ],
                 ],
-                'limit' => 0,
-            ]);
+            ], ['returnScalar' => 'id']);
         return $response->getTotalResults();
     }
 }
--- a/application/src/Api/Representation/ResourceClassRepresentation.php
+++ b/application/src/Api/Representation/ResourceClassRepresentation.php
@@ -23,8 +23,7 @@ class ResourceClassRepresentation extends AbstractVocabularyMemberRepresentation
         $response = $this->getServiceLocator()->get('Omeka\ApiManager')
             ->search('items', [
                 'resource_class_id' => $this->id(),
-                'limit' => 0,
-            ]);
+            ], ['returnScalar' => 'id']);
         return $response->getTotalResults();
     }
 }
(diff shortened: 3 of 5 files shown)


zerocrates:
There's definitely something afoot there... there's broadly no point to select anyactual data in these queries at all, not even the IDs as your change does... we may need to expose an option or interface to explicitly support just counting through the API.

The strange thing is: that's what `limit => 0` is there for already: it's supposed to (and does) actually return zero results. Your changes should be selecting and discarding the IDs of the results... I can't quite figure how that's more efficient. What kind of improvement in performance were you seeing?

zerocrates:
The only thing that I can think of that would account for what you're seeing is the skipping of finalize... but that's trivially fast (especially with zero results), with one possible exception: finalize is where the `execute.post` and `search.post` events are fired, meaning modules can run code at that point.

What kinds of modules are you using? Mapping in particular used to have a "caching" behavior that was used on item searches that rings a bell here. (It was removed in Mapping version 1.1.1).

kloor:
For a property used in 460,183 items, it takes the original method 13 seconds and the new method 10 seconds. We are currently using the CSV Import and RDF Datatype modules.

Doing more testing, it appears that [this group by](https://github.com/omeka/omeka-s/blob/596401a415a11e743276400935051ed3a40a6007/application/src/Api/Adapter/AbstractEntityAdapter.php#L212) is the main culprit. I'm guessing that MySQL may optimize the query if the only field selected is the same as the group by. It doesn't appear that any of the steps after the query make a significant impact.

I quickly made some changes to try to make the query as barebones as possible if the limit is zero. It still takes around 8 seconds to run. Looking at the resulting SQL, I don't really see a way to make it more optimal as it still has to scan every row of that property.

Changing how the search method works based on the limit being set to zero seems odd. Would it make more sense for there to be an additional option specifying that the search is for a count, or to have a separate count method?

zerocrates:
Of course I missed the fact that we _already_ do the optimization I'm talking about: if the limit is set to zero the underlying query doesn't get run at all. So your version is replacing a single query (just the count) with two. And it's still faster?

Anyway... on the subject of the ID restriction, that does allow MySQL to return results directly from the index, since the primary key is the index being used here anyway, so that would make some difference. We don't need to select anything in particular in the inner query to get the count to work, so that's something we could change for counting to just always do (and really something I'd expect Doctrine to do for us as we're using their infrastructure for this counting support but, oh well).

kloor:
I think its one query that is getting run in either case, or at least that's what I've seen from logging the SQL statement. The best I can figure is the non-returnScalar version is not selecting any rows of data because of the zero limit, but it still needs to both join to the value table and scan rows to determine which items have those properties. With the addition of the group-by, I think it has to performing the grouping of all of the data even though it isn't being fetched in the end.

The version in the pull request just selects the ID field, and returns an array of those IDs which then PHP counts. I'm guessing the query runs faster because it only needs to group the one field, but it does seem faster still if it didn't have to group at all, in which case MySQL's count functions can be used instead of PHP's.

zerocrates:
Ah yes, I missed that the count happens in PHP for returnScalar.

You have to have the group by here or an equivalent like DISTINCT because of the joining: you'll get an inaccurate count otherwise whenever any resources have multiple values for the same property.

zerocrates:
Can you try out and share your experience with a branch (or just a patch, it's one small commit): https://github.com/omeka/omeka-s/compare/count-optimization

The idea here would be to keep the API queries as they are but just manually exclude the actual selection of the columns from the subquery when counting.

kloor:
Ah, I didn't consider multiple values for the same property as our dataset currently has none.

That new branch is averaging about 9 seconds for me, which is certainly the best of the versions including the group by.

zerocrates:
Ok so we've merged that branch... the times you're quoting are still _really_ bad, even taking into account a large database, though.

zerocrates:
So I think we won't take this PR... but could transition to just an issue for continuing or other performance concerns.

kloor:
I'm fine with closing this PR in favor of that other branch. I think to a certain extent we're going to face performance issues with the amount of data we're attempting to use, the value table for example currently has 4.7 million rows. I do still need to investigate if I can tune the MySQL server better, perhaps reducing disk I/O. The bigger issue is browsing properties which currently runs this slow query once for each of the 25 rows displayed.

PR #1416, opened 2019-05-21
Extend resource search to titles of linked resources

### LINKED ISSUE #1414: Resource search should extend to titles of linked resources
With resource-titles merged, we now have a much simpler and more accurate method to get "the title" of a resource.

We can use this in the existing search system so that text searches can include results where the value is an internal resource link to another resource which has a title that matches the query.

Especially since we render such linkages by displaying the title when printing metadata for the resource, our current behavior goes against peoples' reasonable expectations.

jimsafley:
Fixes #1414


zerocrates:
Functionally I think we're fine, but performance could be an issue: on a ~30,000-item install the time for a simple search roughly doubles from ~700ms to ~1400. If this is more of in the realm of a constant-ish slowdown it's potentially fine.

I had presumed a join might be faster but it seems to be significantly slower, at least in the "best" case. Possibly because Doctrine creates joins to all the subclass tables when you join to Resource.

The only functional concern I have: the subqueries are currently selecting from Item, this should be Resource.

zerocrates:
On performance there's also the option of making the extension to linked resources an opt-in, though that adds more complexity to an already cluttered search UI so it's probably preferable not to do that.

PR #1421, opened 2019-06-24
Fixed GD thumbnailing of an image with an icc profile.

Daniel-KM:
When an image has an icc profile and gd is used, colors are no kept. This pr fixes it. 
The issue may be the same for imagick, but it depends on the version. A fix for it can be done too, like in https://github.com/Daniel-KM/LibraryDeepzoom/pull/1.

### DIFF
--- a/application/src/File/Thumbnailer/Gd.php
+++ b/application/src/File/Thumbnailer/Gd.php
@@ -28,6 +28,11 @@ class Gd extends AbstractThumbnailer
     protected $origHeight;
 
     /**
+     * @var string The profile icc of the original image.
+     */
+    protected $icc;
+
+    /**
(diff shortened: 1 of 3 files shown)


zerocrates:
So this just copies over the ICC profile from the input to the output; it's not doing any actual color conversion, correct?

Daniel-KM:
yes, it's a simple copy of the profile. You may find another way to extract/copy the profile, but the library works well.

Daniel-KM:
You can try with such an image : https://i.stack.imgur.com/p42W6.jpg

Daniel-KM:
Maybe a conversion may be a better way, but often unperfect.

PR #1422, opened 2019-06-24
Added a link to show current resource in user page.

Daniel-KM:
To finish https://github.com/omeka/omeka-s/pull/1283 with a link to resource show, that is useful when some property are hidden in public side.

### DIFF
--- a/application/src/View/Helper/UserBar.php
+++ b/application/src/View/Helper/UserBar.php
@@ -146,7 +146,7 @@ class UserBar extends AbstractHelper
                 $links[] = [
                     'resource' => $controller,
                     'action' => 'show',
-                    'text' => $translate('Show'),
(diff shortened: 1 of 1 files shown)


Daniel-KM:
Rebased.

zerocrates:
All this is doing is always presenting the View (or Show) link, regardless of whether the user can edit the resource in question, correct?

zerocrates:
I think we'd want to keep the English string as "View" here, even though it doesn't match up with the name of the action... it's consistent with other places like "View Public" on sites admin. Describes what the user is doing rather than what Omeka is doing.

Daniel-KM:
For question 1, there is no check for "view", but rights are checked to display the edit action or not.

PR #1423, opened 2019-06-25
Added an option to list only used properties and resource classes in advanced search form.

Daniel-KM:
A new option is added in site settings : the possibility to display only the used properties and resources classes for the public advanced search form.
Available in module [Next](https://github.com/Daniel-KM/Omeka-S-module-Next) too (version 3.1.2.8) via the trigger `form.vocab_member_select.query` (may avoid to add a key in buildQuery()).


zerocrates:
Closing to focus on #1793 for this feature.

PR #1425, opened 2019-06-27
Add batch update value visibility

### LINKED ISSUE #1424: Batch update value visibility
In the batch update form we should a) add visibility to the "Add * value" elements and b) add a new element to the form that will set all values of a selected property to public or not-public.

jimsafley:
Fixes #1424


zerocrates:
So this adds both a toggle for when you're adding new values as well as a control for modifying the visibility of existing values, yes?

PR #1431, opened 2019-07-23
Fix/item showcase

Daniel-KM:
The key for the item showcase block layout is not the good one, so it is possible to fix it to simplify management of blocks.


zerocrates:
I cherry-picked the header fix; I think we'll just leave the weird ID of the showcase block as it is though.

PR #1432, opened 2019-07-30
Refactor environment check

jimsafley:


application/view/omeka/migrate/index.phtml
```diff
@@ -1,9 +1,13 @@
...
+<p><?php echo $this->translate('There was a recent software upgrade and the database needs to be updated. The entire site will be down for maintenance until you click the button below.'); ?></p>
 <form method="post">
-    <button name="submit"><?php echo $translate('Update database'); ?></button>
+    <button name="submit"<?php echo $environment->isCompatible() ? '' : ' disabled'; ?>><?php echo $this->translate('Update database'); ?></button>
 </form>
+<?php if (!$environment->isCompatible()): ?>
+<p><?php echo $this->translate('There are environment errors that must be resolved before you can update the database.'); ?></p>
+<ul>
+    <?php foreach ($environment->getErrorMessages() as $errorMessage): ?>
+    <li><?php echo $errorMessage; ?></li>
```
zerocrates: should escape here

application/view/omeka/migrate/index.phtml
```diff
@@ -1,9 +1,13 @@
...
-<p><?php echo $translate('There was a recent software upgrade and the database needs to be updated. The
-entire site will be down for maintenance until you click the button below.'); ?></p>
+<h2><?php echo $this->translate('Update database'); ?></h2>
+<p><?php echo $this->translate('There was a recent software upgrade and the database needs to be updated. The entire site will be down for maintenance until you click the button below.'); ?></p>
 <form method="post">
-    <button name="submit"><?php echo $translate('Update database'); ?></button>
+    <button name="submit"<?php echo $environment->isCompatible() ? '' : ' disabled'; ?>><?php echo $this->translate('Update database'); ?></button>
 </form>
+<?php if (!$environment->isCompatible()): ?>
+<p><?php echo $this->translate('There are environment errors that must be resolved before you can update the database.'); ?></p>
```
zerocrates: "environment errors" might be a little opaque but i suppose it's fine

zerocrates: There should be some sort of handle here for styling purposes, a div with a class wrapping around both the paragraph and the list probably?

@kimisgold could have input

application/view/omeka/migrate/index.phtml
```diff
@@ -1,9 +1,13 @@
...
-$translate = $this->plugin('translate');
-?>
-<h2><?php echo $translate('Update database'); ?></h2>
-<p><?php echo $translate('There was a recent software upgrade and the database needs to be updated. The
-entire site will be down for maintenance until you click the button below.'); ?></p>
+<h2><?php echo $this->translate('Update database'); ?></h2>
+<p><?php echo $this->translate('There was a recent software upgrade and the database needs to be updated. The entire site will be down for maintenance until you click the button below.'); ?></p>
 <form method="post">
-    <button name="submit"><?php echo $translate('Update database'); ?></button>
+    <button name="submit"<?php echo $environment->isCompatible() ? '' : ' disabled'; ?>><?php echo $this->translate('Update database'); ?></button>
```
zerocrates: Maybe we just want to not show the form at all here if there's an error

again @kimisgold

application/view/omeka/migrate/index.phtml
```diff
@@ -1,9 +1,13 @@
-<?php
-$translate = $this->plugin('translate');
-?>
-<h2><?php echo $translate('Update database'); ?></h2>
-<p><?php echo $translate('There was a recent software upgrade and the database needs to be updated. The
-entire site will be down for maintenance until you click the button below.'); ?></p>
+<h2><?php echo $this->translate('Update database'); ?></h2>
```
zerocrates: We went through a whole process of adding the `$translate` variables to the views, I don't see much point in taking it out here

application/view/omeka/migrate/index.phtml
jimsafley: Note that the above paragraph says "The entire site will be down for maintenance _until you click the button below_." Iy would be confusing if there was no button.

application/view/omeka/migrate/index.phtml
jimsafley: It's already done. No point putting it back.

application/src/Stdlib/Environment.php
```diff
@@ -0,0 +1,86 @@
...
+     * @var array Environment error messages
+     */
+    protected $errorMessages = [];
+
+    /**
+     * @param ServiceLocatorInterface $services
+     */
+    public function __construct(ServiceLocatorInterface $services)
+    {
+        $translator = $services->get('MvcTranslator');
```
zerocrates: Perhaps we should be using the Message container for these and just translating on output in the installer and upgrader

jimsafley: Why not just translate them here, at the point where they're set?

zerocrates: we mostly translate at print time, largely to avoid doing what this does: injecting the translator "deep" into services

this service could then also be constructed with just the Connection directly rather than taking the service manager, as that's the only service it uses (of course this could be true now but still)

jimsafley: We shouldn't pass the Connection into the constructor because the Environment tests if the Connection factory throws an error. If we pass the connection, the Connection will have already been created (or attempted to be created).

zerocrates: hmm, will the connection actually be created at that call time ever? I want to say it won't... it's actually `connect()` that we're catching errors in, no?

jimsafley: Interesting point. I assumed the installer task fetched the Connection in the `try...catch` because it needed to check if the factory was throwing. If that's not the case (and I suppose it's not) then I'll pass the Connection to the constructor.

zerocrates: I think it's a case where old assumptions were that that _would_ be the case (and maybe it even was in earlier versions... though I doubt it somewhat). But now with the Status service and even moreso the ModuleManager (since it gets the Connection right away before it decides whether or not to try using it), the service will have failed or been created well before the installer or upgrader goes to check it.

application/view/omeka/migrate/index.phtml
jimsafley: Let me be clear: I hate our early use of `$foo = $this->plugin('foo');` at the top of the templates. It's unnecessary, hardly saves any space, and is arguably less readable than simply `$this->foo()`. A few years ago I've slowly transitioned away from setting variables, and removed them when appropriate. I believe this time is appropriate as any.

application/view/omeka/migrate/index.phtml
kimisgold: My main suggestion would be that the list be set to `<ul class="messages">` and each error use the `<li class="error">` markup.

application/view/omeka/migrate/index.phtml
kimisgold: I'm fine with the disabled error button, especially in regards to Jim's point.

PR #1437, opened 2019-08-07
Use of reserved word "value" as a table and column name

julia-bauder:
"Value" is a reserved word in some versions of MySQL/MariaDB. If you're using one of those versions, queries that access or modify the "value" table currently fail with a syntax error. Escaping "value" as a table and column name fixes the error for people running one of those versions, and should not affect usage by people not running one of those versions.


zerocrates:
Can you clarify the versions you've had problems with, if any?

Or perhaps it's some other setting and not the version at issue? We've run S on a decent variety of versions of both MySQL and Maria and haven't encountered this as an issue.

julia-bauder:
10.4.7-MariaDB

zerocrates:
We've had another report of this problem, with the same version... it's still a little curious as I can't find any mention in the documentation or release notes of 10.4 changing the handling of VALUE as reserved. I'm left thinking it's one of: a purposeful change that was undocumented, something else like a changed SQL_MODE (this is reported in the Omeka S system information page), or it's a bug in 10.4.

Anyway, we can quote the usage of `value` as you suggest without issues, but it's just curious. And given we've had 2 reports now close in time to each other after none to this point, it suggests something that was a recent-ish change.

PR #1456, opened 2019-10-02
Added a trigger to finish preparation of values of a resource.

Daniel-KM:
Hi,

There is no possibility currently to reorder the values of a property of a resource. The only possibility is to manage the display of the values, that is useful, but doesn't allow to manage many things. The other possibility is to reorder them at entity levels, but it's complex for the values. So a simple trigger placed in the method values() is userful.

In fact, the module [Internationalisation](https://github.com/Daniel-KM/Omeka-S-module-Internationalisation) needs them : to display the metadata of a resource in a selected language (with managed fallbacks, the simple way is to reorder/filter them. When reordered, the other methods like displayTitle, etc. can be used directly, without change (so it avoids to fix themes). 

The module Internationalisation is working, but not yet finished for an easier management of sites/pages by group of translations.

Sincerely,

### DIFF
--- a/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
+++ b/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
@@ -276,7 +276,7 @@ abstract class AbstractResourceEntityRepresentation extends AbstractEntityRepres
             $values[$term]['values'][] = $value;
         }
 
-        // Order this resource's values according to the template order.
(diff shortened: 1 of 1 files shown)


Daniel-KM:
Note that the module Internationalisation is not yet working on Omeka 2, so the same patch can be applied on Omeka 1.4.

Daniel-KM:
This feature is a step to fix #1456 too.

PR #1458, opened 2019-10-08
Sorting the second-level

symac:
Currently the owners are sorted, but the second level of values appear to be randomized. This PR just adds a ```asort``` before returning the values for the select.


symac:
Whoops, it seems it is something that I want to get fixed, as I had already proposed back in May an equivalent PR ( #1458 ). Closing this a duplicate :)

PR #1463, opened 2019-10-21
[API] adding an option to search media by media_type

symac:
Proposed change based on a discussion [on the forum](https://forum.omeka.org/t/internal-usage-of-the-api-to-search-medias-with-a-specific-mime-type/9652).

### DIFF
--- a/application/src/Api/Adapter/MediaAdapter.php
+++ b/application/src/Api/Adapter/MediaAdapter.php
@@ -19,6 +19,7 @@ class MediaAdapter extends AbstractResourceEntityAdapter
         'created' => 'created',
         'modified' => 'modified',
         'title' => 'title',
+        'media_type' => 'mediaType'
     ];
 
     public function getResourceName()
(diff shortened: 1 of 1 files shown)


Daniel-KM:
Hi,
Note that this feature is already available in the module [Advanced Search](https://github.com/Daniel-KM/Omeka-S-module-AdvancedSearch).

symac:
> Hi,
> Note that this feature is already available in the module [Advanced Search](https://github.com/Daniel-KM/Omeka-S-module-AdvancedSearch).

Thanks @Daniel-KM. I had seen your module but from what I understand, it offers enhancements to the search form, when what I need is a direct option on the API to run a specific search on media_type (from another module : ExtractOCR). So I think both options are interested (but I might have missed something from your module that could meet my need).

zerocrates:
Since the filter doesn't exist in the core, Daniel's module both adds the filter itself to the API and adds the corresponding UI element to the advanced search page.

zerocrates:
@symac there's a very small code style issue with this commit: for multi-line arrays the style is to include a trailing comma after every entry, including the last (so that new items can be added without modifying unrelated earlier lines).

If you could make that small change to include a comma after your added entry to the sorting array, we should have no problem to merge.

symac:
> If you could make that small change to include a comma after your added entry to the sorting array, we should have no problem to merge.

this is now done, thanks for the information regarding Daniel's module, I will have a look as I also need to be able to search by source.

zerocrates:
Thanks

PR #1464, opened 2019-10-21
Feature/mimetype xml zip

Daniel-KM:
Hi, 

In Omeka, all xml files are identified as `text/xml` and zip files as `application/zip`, so it’s not possible to make a distinction between a mets file and an ead file, or to identify an epub, that is a zipped xhtml. This feature is required to use the module [Verovio](https://github.com/Daniel-KM/Omeka-S-module-Verovio) when files don’t use the extension "mei". 

The default list of the xml media-type can be improved for older or newer formats. Note that many xml formats have no registered media type, so they are added to the "vnd" or "x" trees. So maybe they can be added via the config.

### DIFF
--- a/application/src/File/TempFile.php
+++ b/application/src/File/TempFile.php
@@ -363,6 +363,9 @@ class TempFile
         if ($mediaType === 'text/xml' || $mediaType === 'application/xml') {
             $mediaType = $this->getMediaTypeXml() ?: $mediaType;
         }
+        if ($mediaType === 'application/zip') {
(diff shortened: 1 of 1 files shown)


Daniel-KM:
Of course, this feature is already available in module  [Next](https://github.com/Daniel-KM/Omeka-S-module-Next).

Daniel-KM:
This feature has been moved from module [Next](https://gitlab.com/Daniel-KM/Omeka-S-module-Next) to module [XmlViewer](https://gitlab.com/Daniel-KM/Omeka-S-module-XmlViewer).

Issue #1462, opened 2019-10-17
Difficult UI for adding many/all item sets to a site


zerocrates:
This is a close cousin of the issues with the Attachment interface from some time ago: it's pretty clunky to use when you want to do many item set assignments at once, since you have to keep using the sidebar and selecting sets one by one...

Sets also don't get removed from the sidebar when already selected, and the open section gets reset whenever one is picked.

zerocrates:
@mebrett can you take a look over the new workflow here?

It's now on develop.

mebrett:
What's the expected behavior?

kimisgold:
* When selecting an item set, its parent in the right hand selector no longer collapses automatically.
  * Clicking on the owner/parent should still expand and collapse the respective item sets.
* The view scrolls a newly added item set to the top of the page.
* After an item set has been added to the resource pool, the total count and owner count of available item sets updates.
* When removing an item set from a resource pool, the row is removed from the main content area, and the item set becomes available in the right hand selector.
  * The total count and owner count should also update accordingly here.

I think that covers the main behaviors.

mebrett:
New item sets are added at the end of the list of item sets, and the page scrolls a bit so that the item set is near the top of the page. 

Owner counts update as expected, and the owner stays open, even when other owners are open. 

The display of the delete button and the owner's email are overlapping:
<img width="1011" alt="Screen Shot 2019-11-14 at 12 58 37 PM" src="https://user-images.githubusercontent.com/2314070/68883653-7a723480-06df-11ea-9479-23feb167bfdf.png">

kimisgold:
That looks like it's missing some of the new styles—in particular, that sortable handle column should be much narrower. Can you double check with a hard refresh?

mebrett:
Hard refresh resolved the issue with the placement of the delete icon.

zerocrates:
@mebrett what's your feeling on the scroll behavior? It's the same thing as happens when you add a value to the item edit form: we're trying to scroll it to the top, basically. I talked with Kim a little bit about it but didn't really come to a conclusion on whether we thought we should change it (scrolling just enough to bring it into the visible area would be the other option).

mebrett:
[...]

PR #1473, opened 2019-11-07
Revise item set management workflow. (#1462)

kimisgold:
* Keep owner heading in selector open after
  clicking item set.
* Clean up styles for marking item set for 
  removal.
* Hide item set in selector if it exists in the
  item set table.
* Update item set totals in selector after
  adding item sets.


application/asset/js/site-item-sets.js
```diff
@@ -21,46 +28,51 @@ var appendItemSet = function(id, title, email) {
...
-
-itemSets.on('click', '.o-icon-undo', function(e) {
-    e.preventDefault();
-    var undoLink = $(this);
-    var removeLink = undoLink.siblings('.o-icon-delete');
-    var row = undoLink.closest('.site-item-set-row');
-    row.find('input').prop('disabled', false);
+    var row = $(this).closest('.site-item-set-row');
+    var hiddenInput = row.find('input');
+    if (hiddenInput.prop('disabled') !== true) {
```
zerocrates: This whole if/else can just be `hiddenInput.prop('disabled', !hiddenInput.prop('disabled'))`

PR #1479, opened 2019-11-20
Defered js.

Daniel-KM:
To defer js is a quick way to speed a web app (fix #1284).


Daniel-KM:
I kept jquery not defered to remove possible issues in themes. For modules, they are very few that should be fixed (just a wrapper, if it's not possible to create a separate js source).

Daniel-KM:
Rebased for laminas.

PR #1482, opened 2019-12-12
Adapt site user permissions selector to mimic site item set selector.

kimisgold:


application/asset/sass/_screen.scss
```diff
@@ -3465,6 +3466,14 @@ body.sidebar-open #content {
         display: table;
     }
     
+    #user-selector.empty .users-available {
```
zerocrates: Just from an organizational standpoint, presumably this should go in the User Permissions section just below this?

zerocrates:
I'm noticing some remaining styles for the "undo" button that no longer exists... this seems to be the same situation with the item set selector.

PR #1486, opened 2019-12-30
Added a site setting to limit search to some resource types.

Daniel-KM:


application/view/omeka/site/index/search.phtml
```diff
@@ -1,43 +1,51 @@
...
-            true
-        )
-    ); ?>
-    <?php else: ?>
-    <p><?php echo $this->translate('No items found'); ?></p>
-    <?php endif; ?>
+        <?php echo $hyperlink(
+            sprintf($translate('View all results (%s total)'), $result['total']),
+            $url(
+                'site/resource',
```
zerocrates: So basically, this just happens to work because, even though we have a different route for page browse, the URL it resolves to is the same as the one you get if you use the site/resource route and 'page' as the controller? I believe I'm describing the situation correctly.

application/src/Controller/Site/IndexController.php
```diff
@@ -37,29 +37,57 @@ public function indexAction()
     public function searchAction()
     {
         $fulltextQuery = $this->params()->fromQuery('fulltext_search');
+
+        $siteSettings = $this->siteSettings();
+
+        $resourceNames = $siteSettings->get('search_resource_names', ['site_pages', 'items', 'item_sets', 'media']);
```
zerocrates: I think we'd probably prefer if this was defaulted to site_pages and items only (and the same with the default value in the form), so that it's not a behavior change over the current page unless you explicitly go change the setting.

zerocrates:
Other than the note about the defaults, this looks good to me.

PR #1487, opened 2019-12-30
Used skos:definition if available for vocabulary.

Daniel-KM:
In some vocabularies, the meaning of the class or property is set with skos:definition, not rdfs:comment, for example in the skos vocabulary itself. So it can be used like skos:prefLabel is used for the label.


Daniel-KM:
Outdated.

PR #1488, opened 2019-12-31
Added a way to return more than one scalar value from api search.

Daniel-KM:
Allows to get more than one value from the api.


Daniel-KM:
Outdated.

PR #1489, opened 2019-12-31
Returned an array when fetching all values when there is no result.

Daniel-KM:
This is more logic and avoids to check the result when all values are wanted (can be used directly by foreach without setting default).


zerocrates:
This can probably go into a version 3 without issue... otherwise I'd be slightly concerned about the backwards incompatibility.

Daniel-KM:
All pull requests I sent are on my dev server, a demo server. They works fine with all my modules and other ones that I use. I didn't check other modules. For this commit, generally the check is ! or empty(), so there is no issue.

PR #1490, opened 2020-01-01
Adding 'Resource template' column to the list of item and item-set.

fukuchi:
If your archive heavily uses resource templates, you may want to sort your item/item-set list by resource template.

This patch adds the column of 'Resource template' to the list of items/item-sets and allows you to sort the items by the order of the label of the resource template.

A disadvantage of this patch is that the items/item-sets list becomes a little complicated. Here's a screenshot. In the case of our archive, we could hide the 'Class' column, because all items are generated from one of the resource templates.

Untitled

### DIFF
--- a/application/src/Api/Adapter/AbstractResourceEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractResourceEntityAdapter.php
@@ -55,6 +55,18 @@ abstract class AbstractResourceEntityAdapter extends AbstractEntityAdapter imple
             );
         }
 
+        if (isset($query['resource_template_label'])) {
+            $resourceTemplateAlias = $this->createAlias();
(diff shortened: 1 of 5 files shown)


zerocrates:
My initial reaction to this PR was that I'd be more likely to accept a change that makes the columns configurable rather than one that just changes them: I see there's a PR now from @Daniel-KM which takes that approach and I'll look at both of them, but just in terms of approach, I prefer Daniel's version.

fukuchi:
100％ agree with @zerocrates.

zerocrates:
We do now have a "configurable columns" feature on the develop branch, so I think it's best to close this one. Resource template is one of the pre-provided columns you can choose from in the new feature.

PR #1492, opened 2020-01-04
Sanitize the item's description in browse view.

fukuchi:
When you use the default theme and see the item's properties, the texts are sanitized by escapeHtml().
Screenshot from 2020-01-04 14-21-02

But in the list view of the items, the text of the description property is not sanitized.
Screenshot from 2020-01-04 14-20-04

This patch simply solves the problem by encapsulating the text with $escape().
Screenshot from 2020-01-04 14-20-43


zerocrates:
Thanks for your contribution.

PR #1493, opened 2020-01-06
Feature/pretty unescaped json

Daniel-KM:
The [json](https://json.org) standard doesn't require to escape anything, except `"`, "\" and control characters. But php and Zend escape many other characters by default : tags, ampersand, apostrophe and overall the slashes "/" and the unicode characters, making json unreadable in many cases, whereas it's designed to be readable by people and machines. 
So these commits fix this point in the most important cases. There are some other places to fix, but they are useful only for developers, so less important. 
The last commit fixes the api output so it can be readable by people who don't have a json viewer in their browser (and of course it works fine in [Next](https://github.com/Daniel-KM/Omeka-S-module-Next)).

### DIFF
--- a/application/src/Api/Exception/ValidationException.php
+++ b/application/src/Api/Exception/ValidationException.php
@@ -46,7 +46,7 @@ class ValidationException extends BadRequestException
             get_class($this),
             $this->getFile(),
             $this->getLine(),
-            json_encode($this->getErrorStore()->getErrors(), JSON_PRETTY_PRINT),
+            json_encode($this->getErrorStore()->getErrors(), JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_LINE_TERMINATORS | JSON_PARTIAL_OUTPUT_ON_ERROR),
             $this->getTraceAsString()
         );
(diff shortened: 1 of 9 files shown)


Daniel-KM:
Maybe the flat `JSON_UNESCAPED_LINE_TERMINATORS` can be skipped.

zerocrates:
There are definitely some places where we want the default slash-escaping: in particular, in places where the JSON is being rendered inside HTML, or in particular inside a script tag, as that's what prevents JSON containing a closing script tag from ending the element.

Daniel-KM:
I checked the different places, and see that only one place has issue, in job params. This is fixed above. In  fact, the json creation shall be separated from the escaping. For tags included in json, inside a script, it's allowed by the w3c validator.

zerocrates:
The issue with embedding in a script tag isn't about tags or angle brackets, etc. per se, its just about the output containing a closing script tag specifically.

```
<!DOCTYPE html>
<html>
<body>
<script>
<?php echo json_encode('</script><script>alert();</script>', JSON_UNESCAPED_SLASHES); ?>
</script>
</body>
</html>
```

The preceding code will run that `alert`, while the same call without `JSON_UNESCAPED_SLASHES` would not.

PR #1495, opened 2020-01-06
Used the filtered values for the resource title.

Daniel-KM:
Since v2.0, the function `displayTitle()` doesn't use the method `values()` any more, so this value cannot be filtered any more. As it's an important method used nearly anywhere, it should use the same process than other properties. Furthermore, it's now nearly a duplicate of `title()`. 
So this fix reverts the behavior to make it usable like `displayDescription()` and any other `value()`. This fix is required for the module [Internationalisation](https://github.com/Daniel-KM/Omeka-S-module-Internationalisation) and for next improvements of it.

### DIFF
--- a/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
+++ b/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
@@ -488,9 +488,12 @@ abstract class AbstractResourceEntityRepresentation extends AbstractEntityRepres
      */
     public function displayTitle($default = null)
     {
-        $title = $this->title();
-        if (null !== $title) {
-            return $title;
+        $template = $this->resourceTemplate();
+        if ($template && $template->titleProperty()) {
+            $title = $this->value($template->titleProperty()->term());
+            if (null !== $title) {
+                return $title;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


zerocrates:
Part of the idea behind the title change is that the title can be retrieved without loading the set of values, which this eliminates. Using the stored separate title also guarantees that what's displayed is the actual value that's being used for sorting and so on.

I agree that having no filtering ability is an issue, but I'd prefer to just introduce a new title-specific filter, which would also mean it automatically takes care of the shifting of titles between different templates, where if just using `value()`, modules would have to do some more work to make sure they're filtering the property that's being used as the title.

Daniel-KM:
Ok for perfomance. Here, the main point is to make the title translatable in themes: in a site in French, the people want to see the title in French, and Russian people prefer Russian when displayTitle() is used. And of course, the sort should be done with the French or Russian titles, not the latin title if the item  was stored with the title in this language first.
I added the filter in #1506. It works directly for lists and resources views (once the settings are set in the site). Sorting in specified languages (since the module manages fallbacks) is not yet implemented, but it doesn't require a new filter.

PR #1497, opened 2020-01-08
Feature/show template in item list

Daniel-KM:
Based on fukushi improvement #1490, an option to choose the columns to use in the browse views.


zerocrates:
I like this change.

My only hesitation is wanting to enable distinctions between the different browse pages: the one I'm thinking in particular is Media where something like the media type or ingester/renderer really should probably be there, but obviously differences between Items and Sets are probably pretty common in terms of what metadata is actually getting used, too.

Of course this means multiplying out the configuration... I can't immediately think of another way to accomplish it.

zerocrates:
Similar to @fukuchi's, we'll close this in favor of the columns system that we have going into 4.0.0.

PR #1500, opened 2020-01-09
Fixed read of page for home page in user bar.

Daniel-KM:
Sometime, the user bar crashes because there is no page slug.


zerocrates:
What's the situation where this happens? Does it require a module or configuration change to trigger?

Daniel-KM:
I didn't noted the url last month, so I don't remember where it crashes. Probably related to a module, but the fix was in the core.

PR #1502, opened 2020-01-15
Refactoring addAction() and editAction().

fukuchi:
Adds getAddEditView() and make addAction() and editAction() to use it.
Referring this refactoring patch: 01f8dbd


fukuchi:
Sorry, this patch does not work correctly. Later I'll fix this problem. :cry:

PR #1504, opened 2020-01-21
Feature/job step

Daniel-KM:
it improves the jobs with two new values (step and totalSteps). It is useful to follow the progression in real time. In future improvements, it will allow to pause a job too.


flamsens:
This is an excellent idea. Is there a reason why there is no interest to include this in Omeka-S?

PR #1506, opened 2020-01-23
Added event "rep.resource.title" to display a specific title (fix #1495).

Daniel-KM:
Ideally, it should be integrated in version 2.1, so Omeka will support display of translated values automatically through module [Internationalisation](https://github.com/Daniel-KM/Omeka-S-module-Internationalisation) (current dev version use it).


zerocrates:
There shouldn't be a problem squeezing this in

Daniel-KM:
Ok, thanks. In fact, I think it's better to set the filter in function title(), not displayTitle(), so this is the second commit.

PR #1519, opened 2020-03-09
Fixed item set alias for api queries.

Daniel-KM:
When events are used, the query should be reset between calls.

### DIFF
--- a/application/src/Api/Adapter/ItemSetAdapter.php
+++ b/application/src/Api/Adapter/ItemSetAdapter.php
@@ -38,6 +38,8 @@ class ItemSetAdapter extends AbstractResourceEntityAdapter
 
     public function buildQuery(QueryBuilder $qb, array $query)
     {
+        $this->siteItemSetsAlias = null;
+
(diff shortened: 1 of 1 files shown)


zerocrates:
What problem is this causing?

Daniel-KM:
```
Doctrine\ORM\Query\QueryException
[Semantical Error] line 0, col 87 near 'omeka_1.position': Error: 'omeka_1' is not defined.
Details:
Doctrine\ORM\Query\QueryException: SELECT omeka_root FROM Omeka\Entity\ltemSet omeka_root GROUP BY omeka_root.id ORDER BY omeka_1.position ASC, omeka_root.id ASC 
```

zerocrates:
And in what circumstance? Related to events, I suppose? Doing another query inside one of the events for a main query, something along those lines? Anyway, the fix looks correct to me.

Daniel-KM:
It occurs with module Reference and or Search (that can be used alone without Solr) when a list of item sets are used as facets. I didn't investigate more.

PR #1538, opened 2020-04-10
Fix batch user update for site permissions

### LINKED ISSUE #1537: Batch edit user: remove_from_site_permission and add_to_site_permission fields not applied.
Hello,

I am trying to add new site permissions to users using the Batch Edit form but my changes are not applied.

I've seen that the form fields `remove_from_site_permission`, `add_to_site_permission` and `add_to_site_permission_role` are processed in the following function that is itself not referenced anywhere else, although probably intended to be attached to a listener:
https://github.com/omeka/omeka-s/blob/11357cd9f71efe44001210f05e1e680f71532f55/application/Module.php#L409

Also, I was trying to add these fields to the UserForm myself but maybe this would be a nice addition to core. Adding a user permission to multiple sites is a bit of a tedious task today.

Thanks in advance and have a nice day

Laurent

ManOnDaMoon:
Should fix #1537

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -38,7 +38,13 @@ class Module extends AbstractModule
     public function attachListeners(SharedEventManagerInterface $sharedEventManager)
     {
         $sharedEventManager->attach(
-            'Laminas\View\Helper\Navigation\AbstractHelper',
+            'Omeka\Api\Adapter\UserAdapter',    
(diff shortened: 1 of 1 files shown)


zerocrates:
It looks like I managed to remove this event call at some point... probably a mistake when resolving a merge conflict.

PR #1547, opened 2020-04-20
Added displayTitle() and displayDescription() for some entities.

Daniel-KM:
Added displayTitle() and displayDescription() for some entities.


Daniel-KM:
Rebased for laminas.

PR #1551, opened 2020-04-23
Fix unset of raw JSON values from form data

mikesname:
Mistyping of values_json key meant that the encoded values where still present in the returned data after merge.


zerocrates:
Yeah, that's a pretty clear one, huh?

It doesn't cause an actual problem really, but it's obviously wrong.

PR #1554, opened 2020-05-08
Add expand/collapse function to global JS

### LINKED ISSUE #1533: Password Create form : unable to expand field description
Hello

On the password create form (create-password), I'm unable to expand the password constraints hidden in the field description tooltip. A click on the arrow does nothing, although the description is present.
From what I see, collapsible content is handled using admin.js asset, but it is not loaded on this specific form.

Thanks in advance and have a nice day.

Laurent

ManOnDaMoon:
Should fix #1533

### DIFF
--- a/application/asset/js/admin.js
+++ b/application/asset/js/admin.js
@@ -393,20 +393,6 @@ var Omeka = {
             }
         });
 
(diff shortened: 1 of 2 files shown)


zerocrates:
Thanks!

Issue #1553, opened 2020-05-08
Rework .htaccess planned ?


thiboot:
Hello,

I'm preoccupied to serve Omeka S with a maximum security in Apache but the .htaccess packaged is very light. It seems to be a basic ZF3 with just in addition, rules to forbid .ini access and to delete trailing '/'.

https://github.com/omeka/omeka-s/blob/develop/.htaccess.dist

We can access all folders, log file, all others php files independently and errors are not catched to be redirected. Have you a .htaccess more efficient to block direct access to all non-public destinated files/folders ? With rewriting rules more secure ?

In my opinion we should not be able to access php|phtml|phps or log files directly, restrict access to system directories, shouldn't list content of some and correctly redirect erros catched like 403 or 500.

zerocrates:
I'm not opposed to making some of the kinds of changes you're suggesting here...

You're right that the rules are basically the stock zf3 rules with some tweaking. On your concerns of access to php/phtml/phps, replacing the top set of rules with

```
RewriteCond %{REQUEST_FILENAME} -f
RewriteRule !\.(php[0-9]?|phtml|phps)$ - [C]
RewriteRule .* - [L]
```

should be sufficient, causing any bare directory paths or paths directly to php/phtml/phps files to be routed into the application instead. Listing contents of directories is an Apache configuration I don't think we need to mess with, but that same block of rules would cover that problem as well, since directories would be routed to Omeka S.

We could also expand the default deny rules to cover more things like the log folder.

I'm not sure what you're suggesting about redirecting errors, though.

thiboot:
For errors, I'm not sure Omeka catches 403 if we deny access to a subfolder.
I was suggesting to add `ErrorDocument 403 https://my-website.com` for exemple

thiboot:
I'm not familiar with rewriting rules
Replacing
```
# The following rule tells Apache that if the requested filename
# exists, simply serve it.
RewriteCond %{REQUEST_FILENAME} -s [OR]
RewriteCond %{REQUEST_FILENAME} -l [OR]
RewriteCond %{REQUEST_FILENAME} -d
RewriteRule ^.*$ - [L]
```
With
```
RewriteCond %{REQUEST_FILENAME} -f
RewriteRule !\.(php[0-9]?|phtml|phps)$ - [C]
RewriteRule .* - [L]
```
Lead to 500 error :-/
Is there a conflict with following rules ?

zerocrates:
[...]

PR #1556, opened 2020-05-08
feat(config): improve htaccess rewrite rules

thiboot:
close omeka/omeka-s#1553

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -1,26 +1,31 @@
 SetEnv APPLICATION_ENV "production"
 
 RewriteEngine On
+# Uncomment to redirect http to https
+# RewriteCond %{HTTPS} off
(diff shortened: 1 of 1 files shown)


zerocrates:
Apologies again: the version of the .htaccess I'd posted on here was working from a slightly old version, with the result that it rolls back some changes that it shouldn't and in general changes the file more than necessary.

I've committed a corrected version that makes the same changes but leaves things more intact.

You're welcome to submit a PR for your [NC] change which I think is appropriate but I did not yet include, or any other improvements you think would be useful. The commented-out https redirection I don't think is appropriate: that kind of thing in my experience is usually handled at a higher level (server or virtual host configuration, or increasingly through a proxy), and I don't know that we want to muddy the water on that issue.

PR #1558, opened 2020-05-11
Secured dot-files in htaccess.

Daniel-KM:
A complement to #1556 : it protects all dot files, except the directory used by Lets Encrypt (and #1553).

### DIFF
--- a/.htaccess.dist
+++ b/.htaccess.dist
@@ -2,6 +2,9 @@ SetEnv APPLICATION_ENV "production"
 
 RewriteEngine On
 
+# Forbid access to all dot files, except well-known.
+RewriteRule (?:^|/)\.(?!well-known(?:/.*)?$) - [F,NC]
+
 # The following rule tells Apache that if the requested filename
 # exists, simply serve it.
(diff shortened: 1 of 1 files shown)


Daniel-KM:
This is a security issue : without such a point, you can have access to all the history of a web server, potentially discovery hidden data. See https://dev.omeka.org/omeka-s/.git/logs/HEAD. So it's better to set it by default.

zerocrates:
Having things be Git checkouts is common enough that I think I agree that it makes sense to have some default rules for this situation.

I wonder if this is better fitted into the first set of rewrite rules though, same as direct references to php-ish files: so those urls are routed into the application instead of raising Apache-level Forbidden errors. With a leading `!` and an added `[C]` flag I believe it should fit into that stack without issue.

I have to take a minute and blur my eyes to work out the regex here... so it's matching at the root/current or any deeper directory level, directories (and their contents) or files named starting with dots, excluding .well-known, yes?

Daniel-KM:
Indeed, to redirect to an omeka error is probably better looking. 
I found the regex on a site and it works fine, I didn't analyse it more.

PR #1567, opened 2020-05-26
Attached pages

jimsafley:


jimsafley:
@kimisgold For this branch (attached-pages), the new "Site pages" section in application/view/omeka/site/item/show.phtml needs styling.

kimisgold:
In https://github.com/omeka/omeka-s/pull/1567/commits/03e502b6198cb2c0534400558f52c6e377d92c31, I applied the "property" markup pattern to the attached pages section to mirror our treatment of item sets in the show view, as they both describe the use of an item within an installation. I also moved that section above the media list so that it wouldn't get buried under the visual weight of the thumbnails.

PR #1579, opened 2020-06-22
Feature/page site slug

Daniel-KM:
Theme developers needs to search pages by site slug and page slug in views to manage specific pages. And it's a common thing anyway.


Daniel-KM:
Only the last commit is useful, but the rebase keeps all previous commits here for an unknown reason. It simply a way to search page by site slug and not only site id.

zerocrates:
Yeah, I don't know what's going on with this branch... it's got to do with how its targeted at master instead of develop, I think. A rebase and force-push (and maybe changing the target on github) would probably fix it.

But since it's one commit it's easier for me to just cherry-pick it so I went with that route.

PR #1580, opened 2020-06-22
Returned associated array id-value when option "returnScalar" is used in api.

Daniel-KM:
The option "returnScalar" is useful, but it will be more useful when the id is returned with results too. 

Ideally, it may be possible to return multiple selected columns, but for now, the id is enough.

### DIFF
--- a/application/src/Api/Adapter/AbstractEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractEntityAdapter.php
@@ -242,8 +242,8 @@ abstract class AbstractEntityAdapter extends AbstractAdapter implements EntityAd
                     $scalarField, $entityClass
                 ));
             }
-            $qb->select('omeka_root.' . $scalarField);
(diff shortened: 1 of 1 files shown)


zerocrates:
Internally we usually use this as `'returnScalar' => 'id'`... I wonder if there should be some handling to avoid the "redundancy" in that case... but it might not be worth the resulting inconsistency.

Daniel-KM:
I use it for many cases to create quick lists, not only resources. For example list of emails, list of pages slugs, list of resource template labels, etc. So I added an exception for "id", if you need.

zerocrates:
Thinking it over, it's probably best to just always keep the id indexes... it makes it consistent, and sometimes it's useful to have them as the index even if they're also the value, faster lookups from the set and so on.

Daniel-KM:
So you can merge HEAD~1.

zerocrates:
Cherry-picked as 18f5d0be6e46f406ce42c147e03b62598ee3d843

PR #1584, opened 2020-07-01
Allowed to remove a user when there are api keys.

Daniel-KM:


zerocrates:
We already have the Doctrine-level "cascade=remove" on the other end of this relationship which should be handling this.

Anecdotally, I don't have any issue deleting users that have API keys. Are you encountering that as a problem?

Daniel-KM:
It allows direct sql queries too. I have a site with more than 30000 guest users, and each year, many students are removed.

Daniel-KM:
Rebased on develop.

PR #1585, opened 2020-07-01
Feature/iiif api v3

Daniel-KM:
OpenSeadragon has only beta support of iiif api, but it is enough for now.


Daniel-KM:
And another renderer is needed, because iiif v3 allows to display audio and video, not only images.

zerocrates:
This looks good from the spec.

Do you happen to know of any live IIIF Image API 3.0 implementations in the wild?

Daniel-KM:
I was looking, but in fact the module [Image Server](https://github.com/Daniel-KM/Omeka-S-module-ImageServer) I'm finishing is the only one I found.

zerocrates:
OK. I figured they would be hard to come by at this time.

regisrob:
Actually there are a few existing IIIF v3 implementations in the wild, both on the client and server side, and both for the Image and Presentation API. Some institutions also started to expose IIIF v3 contents (Stanford, Oxford, NCSU etc.). 

@zerocrates Is having real IIIF v3 URLs to test with what you were looking for? Or rather other software implementations?

zerocrates:
Just URLs, for the immediate term.

PR #1586, opened 2020-07-06
Added compression in htaccess for text files.

Daniel-KM:
Apache is not compressing by default in some linux distributions, but it increase speed, in particular for api.


Daniel-KM:
Only the last commit is usefull, but github don't see the rebase. 
It simply adds this for performance in default htaccess: 

```htaccess
<IfModule mod_deflate.c>
    <IfModule mod_filter.c>
        AddOutputFilterByType DEFLATE text/plain text/html text/css text/javascript application/javascript application/x-javascript application/ld+json application/json text/xml application/xml
    </IfModule>
</IfModule>
```

PR #1587, opened 2020-07-06
Fix/empty navigation

Daniel-KM:


Daniel-KM:
Rebased.

Only the last two commits are useful, but github doesn't see the rebase.

zerocrates:
There seemed to be something off about the current rebase; I re-rebased it.

PR #1591, opened 2020-07-17
Improvements to process data in bulk with the api endpoint

Daniel-KM:
Added a way to create, update and delete resources in bulk via the api endpoint.

### DIFF
--- a/application/src/Api/Response.php
+++ b/application/src/Api/Response.php
@@ -41,6 +41,7 @@ class Response
     public function setRequest(Request $request)
     {
         $this->request = $request;
+        return $this;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


zerocrates:
My main concern here would be the REST semantics: PUT/DELETE on a list/collection to my understanding implies replacing/deleting the entire collection (which is why Zend calls it replaceList/deleteList, etc.)

Daniel-KM:
Indeed, a post should be one resource. So to keep the rest semantics, it's possible to add some intermediate urls to manage collections, like /api/resource-collection (https://stackoverflow.com/questions/411462/restful-way-to-create-multiple-items-in-one-request#33768160). It's probably not possible to use /api/items for collection because /api/items is already used for a single id (it should have been /api/item/3), unless to change the default api path for that.

Issue #1565, opened 2020-05-26
Resource templates improvements


Daniel-KM:
The resource templates, formerly called input grids in old software, are a main feature of Omeka S, but some points are missing and cannot be added simply by a module. So this is a list of things to work on.

- [x] Multiple data type by property (for example multiple value suggest for subjects and a literal for free subjects; or a creator from different databases). Technically in the Omeka scheme, it can be done with one property with multiple datatypes or multiple properties with one datatype (#1597).
- [x] A specific label for each of the specific datatypes (for example a label Dewey for a normalized subject and a label "Theme" for a value from a custom vocab) (#1597).
- [x] Automatic filling and deletion of the opposite property for value resources (for example when editing item Alpha and adding the property "is part of" to item Beta, set the property "has part" in the item Beta to the item Alpha). Ideally automatically via the the semantic ontology (inverse), else manually via an option in the template (module [Reciprocal](https://github.com/Daniel-KM/Omeka-S-module-Reciprocal)).
- [x] A default value (#1610).
- [x] A non-modifiable automatic value, for example for an identifier, with a specific format (for example a shelf number "mylib-user2-001", or from a helper for more complex formats, like for ark) (#1612, the format of the identifier is managed outside).
- [x] Allow to create a new resource when editing another one for a smoother workflow with linked resources (#1608).
- [ ] Limit the value resources to select to a specific list (with a default query) (postponed).
- [x] Autocompletion with existing values when the user fills a value in a property (#1611).
- [x] Limit the language to some specific ones (more related to ux) (#1610).
- [ ] Qualifying a value (for example the creator can be an author or a director or a producer, or a date can be "uncertain" or "certain", or a place can be a river or a village). This qualification avoids to multiply properties for the same concept, in particular for large libraries. Ideally, it should be done automatically in a semantic way: many properties of various ontologies are related to Dublin Core properties. But it may be done simpler with a merge of multiple properties in forms, display and search. The ux may be similar than the language field, that set a characteristic to the value too; may be a limited list from custom vocab, or a free field.

For these features, it is simpler to upgrade the core, in particular the table resource_template, but the code in many other places too (at least forms, but potentially display, search, api output). 

All changes are in development for August, except the last one.

These features are built for the future [digital library Manioc](http://www.manioc.org) of Université des Antilles and Université de la Guyane, currently managed with [Greenstone](http://www.greenstone.org).

zerocrates:
[...]

PR #1595, opened 2020-07-21
Mutliple datatypes by property in template, and multiple properties with different names

Daniel-KM:
It implements point 1 and 2 of #1565 .

To manage compatibility with current version, it changes the ui only a few, quite unviewable (in resource template form and resource form).  The export is working, but not the import.

In order to manage multiple properties in the same template and compatibility with current version, the duplicated properties must have different data types and label. So it's possible to change template without losing relation between the template.

The database is slightly modified (https://github.com/omeka/omeka-s/commit/2ce1e9e7309f2892f40c8c3175c670a0dd58305a) to manage multiple datatypes. The migration works, but it is not launched because the version number of omeka is not updated in the pr, so you may have to run it manually, or install a full omeka. The sql query is reversible, so you won't lose anything. If somebody needs to check on a living server, I can open an access on a demo.

For the point 3 and next ones (in next pr), new settings should be added to resource template and resource template properties (list of language, reciprocity table, max values, etc.).  It can be a single column settings or one column for each settings. My developments use the first solution (one column array_json) for all new settings, but it may be changed later.


Daniel-KM:
The same, the column "data_type" can be changed into a json_array if needed.

Daniel-KM:
image

application/data/install/schema.sql
```diff
@@ -185,11 +185,12 @@ CREATE TABLE `resource_template_property` (
   `alternate_label` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   `alternate_comment` longtext COLLATE utf8mb4_unicode_ci,
   `position` int(11) DEFAULT NULL,
-  `data_type` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `data_type` varchar(766) COLLATE utf8mb4_unicode_ci DEFAULT NULL,
```
zerocrates, line 188: Pretty sure this is going to be a compatibility issue: not all MySQL installs will allow such a long value and associated index.

Daniel-KM: I didn't check mysql, but it is not a limit for Travis. Anyway, the index can be removed, because it is not really used (a check is done in the adapter).

zerocrates:
This is a pretty significant change and I haven't looked it over in detail.

As an initial note, we don't do any commits directly to the master branch, and don't accept PRs to it either. For some of your smaller previous ones, I've just cherry-picked to avoid any issues, but I wouldn't do that for a larger PR like this one.

Daniel-KM:
> As an initial note, we don't do any commits directly to the master branch, and don't accept PRs to it either.

So where can i add commit or pr?

zerocrates:
As usual we still accept PRs on develop.

Daniel-KM:
Yes, I can rebase it on develop, but It is used on a Omeka 2 version.

Daniel-KM:
Change are not so big: 
- 1-5: some small improvements
- 6-12: implement backend for template with multiple properties with different label and data type
- 13-15: update of resource-form.js to manage multiple properties
- 16-18: implement backend for template with properties with multiple data types
- 19:  update of resource-form.js  to manage multiple data types
- 20-22: update of code to display new templates with resource-values
- 23-25: fix and improvements
- 26: export

Daniel-KM:
I rebased on develop: #1597.

application/data/install/schema.sql
Daniel-KM: This is the default limit of mysql: https://dev.mysql.com/doc/refman/5.6/en/innodb-limits.html, but I'm removing the constraint.

Daniel-KM:
Last commits fix some issues (and the same for the laminas branch). There are no more check on the number of data types of the label, so the librarian can create a template with any property and data-types (but without duplicating data types for the same property). `data_type` is now a json array. Import and export is working too.

Daniel-KM:
The feature "multiple data types by property" is managed via pr #1634 and the feature "multiple labels for property" will be managed directly by the module [Advanced Resource Template](https://gitlab.com/Daniel-KM/Omeka-S-module-AdvancedResourceTemplate).

PR #1599, opened 2020-07-24
Make new media sortable

### LINKED ISSUE #1598: New media are not sortable
Hello

When adding new media to an item, the added sub-forms are not sortable. Users have to add media, save resource, and edit resource again to sort the added media items.

It looks like the cause is only missing element and CSS classes in the `media-field-wrapper.phtml`  view: adding the sort handle and making the corresponding div sortable seems to fix this issue.

ManOnDaMoon:
Should fix #1598


zerocrates:
Works great! One wonders why we didn't do it this way in the first place since the API supported it already.

PR #1600, opened 2020-07-27
Updated easyrdf with upstream.

Daniel-KM:


zerocrates:
Ah, there's a new EasyRDF release, great.

Issue #1602, opened 2020-07-28
Site page feature requests


mebrett:
Two features which would be nice to have, from a project management standpoint (based on work on HtA). The use case is a project which is using Omeka S to build out a site, with multiple contributors, and which is continuing to add complex page content after the site has gone live/public. 

1. Adding a column for the date the page was created in the table at `site/s/sitename/page`, which allows project managers and site admins to quickly see what is the newest page added by a team member.
2. Ability to set a page either not public or to exclude it from site search. Useful for drafts of new content which aren't quite ready to go, being added to an already live site. At the moment, the only way (that I can think of) a site visitor would find these pages is through search.

sharonmleon:
Related to issue #1549, which ask for edit date and issue #1361, which asks for a public/private check box for pages so that they can be reviewed by logged-in users in the public interface before going public.

sharonmleon:
I don't know that we want to close this yet. Accumulating interest, etc.

Daniel-KM:
The second point is fixed in #1603.

Daniel-KM:
I added a date column in #1615.

Issue #1361, opened 2019-01-25
Add a Public/Private check for site pages


sharonmleon:
To facilitate the review of work within the design context, rather the admin interface.

Daniel-KM:
This is fixed in #1603.

PR #1603, opened 2020-08-03
Page visibility

Daniel-KM:
This is a long time requested feature, available in Omeka 2 since the origin.
It fixes point 2 of #1602 and #1361.
To make it working, you need to increase the version number of Omeka (3.0.0-alpha.2) in [application/Module.php](https://github.com/omeka/omeka-s/blob/develop/application/Module.php#L17) if not installed from scratch.


zerocrates:
This is certainly a requested feature.

My concern with this PR is that I don't believe it will take care of correctly excluding private pages from "browse" contexts, like the public page browse/search page. Just marking the fulltext records as private as you're doing here should cover actual searches, but you can get a list of all pages by doing no query. A similar issue would apply to ItemRepresentation::sitePages (new in 3.0, lists the pages an item is attached to). Typically we want to exclude private results from these kind of "browse" areas completely, and do so at the query level so things like pagination still work correctly.

Daniel-KM:
Yes, I forgot this point. I'll update it in a few days.

Daniel-KM:
Here is the update.
There is no owner currently for pages. I have no requests for that, so it is probably useless, but maybe in the future.
For site pages, the block "Browse preview" and some others are not taken in account in the result, only pages with individually assigned items, but this is unchanged.

application/src/Permissions/Assertion/OwnsSiteAssertion.php
```diff
@@ -0,0 +1,20 @@
+<?php
+namespace Omeka\Permissions\Assertion;
+
+use Omeka\Entity\SitePage;
+use Laminas\Permissions\Acl\Acl;
+use Laminas\Permissions\Acl\Assertion\AssertionInterface;
+use Laminas\Permissions\Acl\Resource\ResourceInterface;
+use Laminas\Permissions\Acl\Role\RoleInterface;
+
+class OwnsSiteAssertion implements AssertionInterface
```
zerocrates, line 10: Was there an issue with the previous way this was working with OwnsEntityAssertion in your first version of this PR? Since SitePage has getOwner implemented (as a passthrough to the containing Site's getOwner) it seems to me that it should work fine without needing a one-off assertion like this, and I think it's preferable to have it work that way.

zerocrates:
We're aware of the limitation with the way the "this item appears on these pages" feature interacts with browse preview and other non "attachment" based blocks, and it's totally expected that this change would leave that limitation untouched. So no problem there.

The SQL filter approach is in line with the way we generally handle this issue elsewhere, so I think we should be on solid footing here. We'll have to do a little manual testing and so on to confirm this is working as expected in combination with the various places pages could appear.

application/src/Permissions/Assertion/OwnsSiteAssertion.php
Daniel-KM: Too much in advance.
I removed it.

zerocrates:
Megan, this PR is available for review as the local branch `site-visibility`, it includes just a couple small changes.

mebrett:
First problem: needs to fail more gracefully.
1. On the public side - if a non-logged in user gets a link to a private page (in this case, I pasted from a browser where I'm logged in), they get a Not Found message; if the site is set to development, they'll get a whole error message. If it's possible to give site visitors a nicer landing. Even just tweaking the message to "*Page* Not Found" would help. 

2. A logged-in user who does not have permissions for that site can see that the private pages exist, and can click on the view page link. They get this error: `Omeka\Api\Exception\PermissionDeniedException Permission denied for the current user to read the Omeka\Entity\SitePage resource.`  If it's possible to deactivate the 'view' button for users without permission, it could be useful. 
Screen Shot 2020-08-19 at 4 37 59 PM

zerocrates:
That number 1 is expected: that's just how the not found page looks, we've never really spent time on it visually.

Number 2 is a bug. What's supposed to happen is that the private pages are just left out of that browse list completely if you aren't allowed to see them.

Daniel-KM:
For 2, it's not an issue, it is a non-implemented feature… Currently, it manages visibility for anonymous people, not for users, so the rights for them should be defined when they have / don't have a site permission. 
So:
- Should we distinct each site permission level (viewer/editor/admin) or manage them as a whole (with/without any permission), as currently?
- Should reviewer/author/researcher and editor with a site permission able to see private page (currently yes).
- Should reviewer/author/researcher and editor without site permission able to see private page (currently no, but editor who can see the title). 
I pushed a first fix on this point, but it's mainly about new permissions rules.

zerocrates:
When I said it was a "bug," my view is this: the permissions check for the "browse" for pages should be exactly parallel to the one for viewing a specific page: so whenever a user does not have permission to view the page, it's simply not included in the browse list.

The issue here is the "view-all" permission being granted to editors. I see two valid situations here, either:

1. Editors should have permission to see all pages regardless of visibility even on sites they do not have special permissions assigned for (in which case there should be an additional un-asserted grant of the `read` permission for the editor role, just as it [already has one for update and delete](https://github.com/omeka/omeka-s/blob/27605736b5463e002fb5946669cdc4158bb44aa0/application/src/Service/AclFactory.php#L210). With this setup it would be unnecessary to have the check Daniel just added on the "view in public" link in his latest commits.

2. Editors should _not_ have permission to see private pages in sites they don't have special permission for, in which case they should simply not have the 'view-all' permission granted to them.

I believe, from the way the roles are currently laid out, that option 1 is the most consistent. @mebrett, what do you think?

@Daniel-KM, on your other points you raised:

- For this feature it should be sufficient to treat site permissions all as one, given that we're talking about "viewing," the lowest grantable level. So I don't think that needs to change at this point.
- I think your questions/answers about reviewer/author/researcher are correct: they _should_ be able to see the private pages if they have permissions on the site or own it, and they should _not_ be able to see them if they do not. The only point of contention I have is on editors.

zerocrates:
I made a small mistake in my previous comment, I edited it to remove it.

The situation remains the same: I think we either want to add an unconditional read permission for editors, or remove the view-all for editors. I haven't yet checked, but this situation where the editors can browse but not read feels likely to cause other issues elsewhere, in particular in the API.

Daniel-KM:
So it's mainly about removing the rule I added for editor.

zerocrates:
Yes, we talked it over a little on this side of things and the consensus is, just don't give special permissions to the global Editor role here. Leaves individual sites more freedom to make the decision of who can see their private pages themselves.

So just removing the view-all rule for editor should resolve that.

Daniel-KM:
Ok, it's committed.

zerocrates:
Merged (with some rebasing)

PR #1607, opened 2020-08-24
Allowed to get associated fields when returning scalar from api.

Daniel-KM:
When using the api search, it's not possible to return the scalar values for the foreign key of a table. For example, it's not possible to get a list of user ids of items.

### DIFF
--- a/application/src/Api/Adapter/AbstractEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractEntityAdapter.php
@@ -263,14 +263,21 @@ abstract class AbstractEntityAdapter extends AbstractAdapter implements EntityAd
 
         $scalarField = $request->getOption('returnScalar');
         if ($scalarField) {
-            $fieldNames = $this->getEntityManager()->getClassMetadata($entityClass)->getFieldNames();
+            $classMetadata = $this->getEntityManager()->getClassMetadata($entityClass);
+            $fieldNames = $classMetadata->getFieldNames();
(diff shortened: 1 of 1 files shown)


application/src/Api/Adapter/AbstractEntityAdapter.php
```diff
@@ -235,14 +235,21 @@ public function search(Request $request)
...
-                    $scalarField, $entityClass
-                ));
+                $associationNames = $classMetadata->getAssociationNames();
+                if (!in_array($scalarField, $associationNames)) {
+                    throw new Exception\BadRequestException(sprintf(
+                        $this->getTranslator()->translate('The "%1$s" field is not available in the %2$s entity class.'),
+                        $scalarField, $entityClass
+                    ));
+                }
+                $qb->join('omeka_root.' . $scalarField, $scalarField . '_');
```
zerocrates, line 248: I think it's better off using `createAlias()` for this join's alias as is usual for other joins.

zerocrates:
To my reading this _only_ works on the ID column of associated entities, correct?

I'm wondering if we can do this with the `IDENTITY()` DQL function and avoid the join. It just would require finding only the associations with the correct "owning side," I think.

Daniel-KM:
Indeed, the join is automatically added by doctrine itself with identity().

Daniel-KM:
Yes, the it's only for related entity ids, other data are available with the getFieldNames().

PR #1608, opened 2020-08-24
Allow to create a new resource when editing another one for a smoother workflow with linked resources

Daniel-KM:
Something that is common when creating semantic items, with many relations, is the need to create new items during edition. For example, the librarian is creating a record for a new book and he wants to fill "Dublin Core : Creator" with an author whose record does not exist yet. Currently, it is impossible, he should give up the current item and create the author first. Of course, he can open a new tab, but this is not a smooth workflow. 

So this commit add a new button in the right sidebar, so when he doesn't find an item, he can create it directly.

It improves the resource form too (see #1565 for more improvements).

This feature is needed for the future [digital library Manioc](http://www.manioc.org) of Université des Antilles and Université de la Guyane, currently managed with [Greenstone](http://www.greenstone.org).


Daniel-KM:
This feature is included in the new module [AdvancedResourceTemplate](https://gitlab.com/Daniel-KM/Omeka-S-module-AdvancedResourceTemplate).

PR #1609, opened 2020-08-26
Improved resource templates to manage settings (Laminas)

Daniel-KM:
Based on #1597, it's a next step to manage resource template improvements (#1565). 

In fact, I need to know if you will merge this pr or not. Of course, the core should not have all features, in particular when they can be provided by a module. But to manage multiple data types by property cannot be done in a module and, for this pr, to manage settings in the core will allow to manage many settings in the modules, like #1610 that allows to display a limited list of languages for the values, that will help users.


Daniel-KM:
Rebased directly on develop in #1614.

PR #1610, opened 2020-08-26
Add a select for languages in the resource form

Daniel-KM:
Add a select for languages in the resource form, so the user just clicks a language to set it, so a lot simpler and cleaner. It is done for the resource template improvements (#1565) and is based on #1609.


Daniel-KM:
The conflict is only related to style.css, but I can't run gulp css to fix it here.

zerocrates:
I think the feature of having a pre-set list of in-use language codes is a great one. I have a little quibbling over the select box styling taking up so much space but that's not a major issue, we can always deal with that.

My thinking is that the average user works with a relatively small set of languages... I'm wondering if this makes more sense as a user setting (or possibly a global instead of or in addition to that), rather than a template one. So if you have an installation that's in 3 or 5 or 10 languages, you don't need to go set the list of possible languages on every property separately, you just configure it once.

It also might be a perfect use case for the `datalist` attribute instead of a select, so the pre-entered options are there but the user is still free to vary from them if necessary.

Daniel-KM:
Yes, I have the same issue with the reciprocal properties, initially, it was designed to be a template property setting, but finally it is a global one. The same for many settings in fact: interface language, number of results in browse lists, etc. 

In some other cases, it is not related to the user, but to the institution (when install is multi site, there are many libraries and the items are specific to each library). But the user is related to an institution anyway.

So it's mainly a question of priority: global setting < user settings < templates or global < template < user?

For language, it's interesting to let free entry because sometimes in library or in archives we get an original paper, so I'm fixing it.

Daniel-KM:
.Note that in this pr, there is a global option for each template, then a specific option for each property. It allows in particular to forbid a language for some properties (identifier has no language, etc.). The global or user config can be only generic (a list of languages), but there a need for some more precise config in the template.

Daniel-KM:

PR #1611, opened 2020-08-28
Resource form value auto-completion with existing data, according to resource templates

Daniel-KM:
It implements the auto-completion of the resource form literal values. 
It is built for the resource template improvements (#1565) and is based on #1609 and #1610. It will be rebased on develop or externalised into a module when the point on the resource template settings will be clarified.


Daniel-KM:

PR #1612, opened 2020-08-28
Lock specific values in the resource form once they are saved

Daniel-KM:
It implements the lock of original values in the resource form once they are saved.
It is built for the resource template improvements (#1565) and is based on #1609 and #1610 and #1611.


Daniel-KM:

PR #1614, opened 2020-08-28
Add settings for resource templates

Daniel-KM:
A shorter version of the pull request #1609 for resource template settings, not based on multi-data types improvements (see #1597), so it can be used as a base for improvements of the resource templates by module (#1565).


zerocrates:
So, this is basically just creating storage for whatever "extra settings" modules might want to add for templates and template properties? What kinds of things are you thinking of doing with it?

It's certainly easier and faster to get something like this into the core and then let modules play around with specific new features, at least to start, so I'm in favor of the strategy.

I know this is already a "smaller" version, but can you give me a PR that's just this main feature, without the unrelated JS cleanup and sorting and extra import/export validation and stuff like that? Those could all be their own PRs; it's just much easier to manage on this end when they're separated.

Daniel-KM:
Yes, it's a shorter version and based directly on develop. Except the last one, all commits are related to the resource template settings management. It's not only to create a column for settings in the table, but the resource template form should manage it, so it requires update to the resource template controller, and to the js ; then these new settings should be used in the resource edition page, so the form should be updated, but mainly the resource-form.js. Some commits can be pushed separately, but they are a few.

So I'm going to open multiple pull requests, but each one will be based one on the previous one.

To try this feature directly, you can use the module [Advanced Resource Template](https://gitlab.com/Daniel-KM/Omeka-S-module-AdvancedResourceTemplate) that uses this pr and that add the features I pushed previously and that I removed: autocompletion (#1611), locked value (#1612), language selection (#1610) for the improvements related to resource templates (#1565).

Daniel-KM:
I reset this pr as a merge of the other prs.

zerocrates:
OK here's where we are as a whole for this, might as well comment here for simplicity's sake:

- I've cherry-picked out part of the "unrelated" changes for template import/export of pre-existing features, so that can be removed from this PR
- The language global setting change, it's much appreciated and I'll review it on its own; it shouldn't be included here
- I've made a few specific comments on the various "separated" PRs
- The basic change of adding these columns and pulling from them into the resource form seems good to me
- The specific changes you've made to the resource template form I'm less sure on, there seems to be a fair bit of complexity there that I'm not totally confident of and some simpler changes I don't think I'm in favor of (like the "extra settings" fieldset always being added). Parts of this like the enabling of events for that form are obviously fine, though.
- We do have a theoretically small ask that in practice really touches all of this: can we call the columns here `data`? Other similar directly-on-an-entity columns for this purpose are called `data` already, and the naming reflects a possibly broader scope for what could be stored here beyond just "settings" (if doing this renaming I think we'd prefer to just change the existing migrations, not add additional ones)
- If, as you've indicated elsewhere, the minimal required form of this change is just the database/entity/migration part to actually add the columns and methods, and the piece that makes the data available in the resource form, just _that_ stuff together should perhaps be the PR we move forward with at least to start, if just avoiding the questions about the template form for now simplifies things

Daniel-KM:
> * The specific changes you've made to the resource template form I'm less sure on, there seems to be a fair bit of complexity there that I'm not totally confident of and some simpler changes I don't think I'm in favor of (like the "extra settings" fieldset always being added). Parts of this like the enabling of events for that form are obviously fine, though.

The main idea of this change is to be able to use filters and validators of any element of the form, included property ones. It's a security too (the data should not be the post). I can simplify some points anyway and remove the extra settings.

> * We do have a theoretically small ask that in practice really touches all of this: can we call the columns here `data`? Other similar directly-on-an-entity columns for this purpose are called `data` already, and the naming reflects a possibly broader scope for what could be stored here beyond just "settings" (if doing this renaming I think we'd prefer to just change the existing migrations, not add additional ones)

Ok, i'm going to rebase the pr.

> * If, as you've indicated elsewhere, the minimal required form of this change is just the database/entity/migration part to actually add the columns and methods, and the piece that makes the data available in the resource form, just _that_ stuff together should perhaps be the PR we move forward with at least to start, if just avoiding the questions about the template form for now simplifies things

I published a new feature in the new module [AdvancedResourceTemplate](https://gitlab.com/Daniel-KM/Omeka-S-module-AdvancedResourceTemplate) and a simplified [backport](https://github.com/Daniel-KM/Omeka-S/commits/backport) for version 2 of all the issues, so it's now possible to prefill a resource with an external resource. So I think that resource templates can manage all cases now. The needs are only a management of the settings, events and filters by the resource template form, a js improvement to manage settings  in resource form and in resource template form.

Daniel-KM:
This is the new simplified advanced settings pr. It manages only the resource template part; for the resource part, see #1623.

zerocrates:
If I'm reading this correctly, the sidebar is still just using the preexisting manually-written-out elements, and not the ones defined in the new Form class stuff here.

The result is a lot of changes to the "old" parts of this form and stuff like a copy of the data type select logic so it can be used in the Form class, but none of that's really used for much here: the new "sidebar" elements are all actually rendered, but inside a hidden div in the property rows area (so we have things like hidden elements and then also CSS-hidden checkboxes for the same value, even though it's really being used like a hidden input).

I'm not really in favor of either of those things: the CSS-hidden elements add bloat to the page and will make it easier to grab the "wrong" element in JS, and just in general there's quite a bit of added code here for things like labels and IDs and so on to the form that isn't really being used since the visible form is still manual .

Converting the template to use the Zend/Laminas form system more fully may be a reasonable idea (personally I tend to think it ends up as more trouble than it's worth once the form is dynamic), but I think it would have to go further than what's here, and probably would be a separate issue on its own.

I see a couple options: one would be to just not mess with the form at all at this stage, and make it do a partial update so the o:data is just not messed with by this form/controller at all. Modules could handle modifying the settings on their own, as one possibility (though forms for messing with template properties are somewhat complex so I understand that's not necessarily ideal).

Another would be to just remove the form bits and changes that deal with the existing template data, so leave their previous hidden inputs and so on in place, only adding the parts actually used by the new o:data feature. So Laminas form filtering and so on could still be used, but only for the elements already handled by the existing form class and also for things added by modules to store in o:data.

You might consider, for example, a simple Laminas\Form (or fieldset) that's just for the property o:data alone, used to display the elements in the sidebar. The JS would just serialize the values for that data over to a single hidden JSON-encoded `o:data` element in the "property rows" area of the form (multiple hidden inputs would also be possible for this use but I feel a single is just simpler), and the controller could then just decode that JSON, pass it through your o:data-only form for validation/filtering, and have the results from the form placed in the `o:data` key for the property for sending to the API.

This seems like it would avoid the need for multiple "fix" methods and the duplication of elements that's present in this current version.

zerocrates:
Just a heads-up also that we're looking to close the window on changes going into 3.0 soon, like the end of this week, to allow some down time for people to update modules and so on.

To my mind there'd be no real issue having this or other template stuff in a 3.1 release if that were necessary, but I thought I'd let you know.

Daniel-KM:
And what about the multiple datatype improvement?

zerocrates:
I have some lingering concerns on it (#1597) also:

- It looks to currently contain some pieces of this PR (the obvious bit is a bunch of JS looking for `setting-key` attributes), which should be removed
- It currently doesn't apply cleanly, though the resolution is simple
- I'd appreciate some paring down of the 37 commits it currently contains, a simple rebase would probably deal with this and the previous point
- There's some oddities with how it functions: the Default "type" is selectable and appears as a choice of type even though it's really the "blank" choice (it should perhaps be removed entirely from that select, or made the "empty option"
- Issues of style and visual problems like that are around a bit: missing padding between the multiple data type choices in the resource form, problems with correctly displaying the "Title" and "Description" check icons when there are multiple templates for that property...

More generally and probably importantly, in our internal discussions, we're comfortable with the multiple data types improvement; it does affect one module we know of but that's not a terribly major issue.

We're less certain on the "multiple template entries for a single property" part: there's a fair amount of extra complexity there, the PR ends up having to touch the value retrieval in the representation, and issues specific to it such as dealing with the number of "fields" for a property shrinking or growing when switching templates (something it still can't quite handle doing "live," I believe), and things like where values "should" go, say if there's one template property with the Default type and one with a specific one: the code is clear on what it will do but this creates a situation where the data entry UI doesn't reflect what's actually happening.

As I believe I mentioned or suggested before, we could deal with those pieces separately: in particular, the multiple data types implementation alone: that's just, if my understanding is correct:
- the alteration of the database storage for the template properties
- updated methods on the entity/representation/API for template properties to save/retrieve those multiple values
- a very small change in the template form to just switch the data type input to multiple and save the data correctly in the "real" hidden inputs
- the changes to the resource form itself to handle the UI for multiple types

That would be a change that's much more constrained in what it affects and so we'd be more confident in applying it.

If the timing doesn't work out, my thinking is pretty much the same here as for the other template features: I think they could be shifted to 3.1 if necessary.

Daniel-KM:
For the settings, I'll publish a new version of the module [Advanced Resource Template]() based on develop with many more js in order to override features. The point is the resource form js in #1623.
For the datatype, I have rebased the basic version with multiple data types here #1634.

Daniel-KM:
For the "multiple template entries for a single property", it can be implemented via a specific setting in rows (a list of data types and labels), so it will be available in the module Advanced Resource Template.

Daniel-KM:

PR #1615, opened 2020-08-28
Added a column to order pages by date

### LINKED ISSUE #1549: Sites: Pages: Browse -- add a column that contains most recent edit date
Would be nice to see most recent edit date stamp on the pages browse table. Seems like there is plenty of room to the right of the slug column.

Daniel-KM:
Fixes #1549 #1602.


zerocrates:
Looks good to me.

My only question: I'm wondering if we should skip displaying the indentations if not using the "nav" order, so we're not implying parent/child relationships that don't exist with the alternative orders.

PR #1617, opened 2020-09-02
Added a main setting to list suggested languages for values.

Daniel-KM:
According to https://github.com/omeka/omeka-s/pull/1610#issuecomment-681114762, it adds a new main setting to add a select (datalist) for languages in resource form.

### DIFF
--- a/application/src/Form/Element/ArrayTextarea.php
+++ b/application/src/Form/Element/ArrayTextarea.php
@@ -101,14 +101,14 @@ class ArrayTextarea extends Textarea implements InputProviderInterface
     }
 
     /**
-     * Get each unique line of a string separately as a list.
+     * Get each line of a string separately as a list.
      *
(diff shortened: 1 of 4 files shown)


application/src/Form/Element/ArrayTextarea.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * Get the option as key/value or simple list.
+     *
+     * @return bool
+     */
+    public function getAsKeyValue()
+    {
+        return $this->restoreValue;
```
zerocrates, line 145: should this be `asKeyValue` ?

Daniel-KM: Indeed. Initially, I thought to use the RestoreTextarea datatype, but finally not.

PR #1619, opened 2020-09-02
Core part to manage resource template settings

Daniel-KM:
This is the core part to manage resource template settings (see https://github.com/omeka/omeka-s/pull/1614#issuecomment-683904981).


zerocrates:
Thank you for splitting this up. It may seem a little silly, but it really is much easier for me to manage things this way.

For reference, my concern was more the "extras" in terms of fixes and features, like the sort order of the templates and including the already-existing title and description properties in the import/export for templates. Having those kinds of things separate makes review easier, and makes it easier to create an accurate changelog.

The actual "settings" feature you're adding with these, that is fine and in fact preferable as a single PR that contains the database changes, the UI changes, the import/export changes, etc., just as long as it's only the things that directly concern the feature. The extra splitting you've done here is fine, but I just wanted to clarify why I asked you to change your previous PR.

Daniel-KM:

PR #1621, opened 2020-09-02
New forms to manage the resource template form with settings.

Daniel-KM:
In order to manage settings (#1619), the resource template form should be event aware (to let modules add new settings) and should manage a collection of resource template property fieldsets, that should be event aware too (let modules add specific ones for properties). 

Because it uses the Zend element Collection for properties, the filtering and  the validation of elements (internal or added by modules) of the property fieldsets are simpler if they are real elements, so a select is added for datatype too.

### DIFF
--- a/application/src/View/Helper/DataType.php
+++ b/application/src/View/Helper/DataType.php
@@ -82,10 +82,12 @@ class DataType extends AbstractHelper
     {
         $view = $this->getView();
         $templates = '';
+        $resource = isset($view->resource) ? $view->resource : null;
+        $partial = $view->plugin('partial');
         foreach ($this->dataTypes as $dataType) {
-            $templates .= $view->partial('common/data-type-wrapper', [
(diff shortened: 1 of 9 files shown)


application/src/View/Helper/DataType.php
```diff
@@ -74,6 +74,9 @@ public function getSelect($name, $value = null, $attributes = [])
         $element->setEmptyOption('Default')
             ->setValueOptions($options)
             ->setAttributes($attributes);
+        if (!$element->getAttribute('multiple') && is_array($value)) {
```
zerocrates, line 77: Is this stray stuff from the multiple datatypes change?

Daniel-KM: It's separated : as long as there is a data type helper, it should manage a form with single and multiple data types, somewhere else than in the resource template form. But indeed, it is needed for the multiple data types.

Daniel-KM:

PR #1622, opened 2020-09-02
Manage resource template settings in the admin board

Daniel-KM:
Based on the new feature #1619, it updates the resource template controller and the associated views in order to manage new resource template settings {#1614). Of course it uses the new forms (#1621).

PR #1623, opened 2020-09-02
Use resource template settings in resource edit pages

Daniel-KM:
In fact, only the js is updated to make the settings working in resource edit pages, so this pr is not based on #1622. So it cleans and factorizes some parts of the code in order to be able to apply the settings in all cases, and add a new trigger "o:template-applied" in order to be able to let modules to know when a template is applied. It can be merged with the event "o:form-loaded" if needed, because to load or remove a template is nearly the same.

### DIFF
--- a/application/asset/js/resource-form.js
+++ b/application/asset/js/resource-form.js
@@ -25,7 +25,7 @@
         $('a.value-language').on('click', function(e) {
             e.preventDefault();
             var languageButton = $(this);
-            var languageInput =  languageButton.next('input.value-language');
+            var languageInput = languageButton.next('input.value-language');
             languageButton.toggleClass('active');
             languageInput.toggleClass('active');
             if (languageInput.hasClass('active')) {
@@ -37,7 +37,7 @@
             if ('' === this.value || Omeka.langIsValid(this.value)) {
                 this.setCustomValidity('');
             } else {
-                this.setCustomValidity(Omeka.jsTranslate('Please enter a valid language tag'))
+                this.setCustomValidity(Omeka.jsTranslate('Please enter a valid language tag'));
             }
         });
 
@@ -46,7 +46,7 @@
             e.preventDefault();
             var typeButton = $(this);
             var field = typeButton.closest('.resource-values.field');
-            var value = makeNewValue(field.data('property-term'), null, typeButton.data('type'))
+            var value = makeNewValue(field.data('property-term'), typeButton.data('type'));
             field.find('.values').append(value);
         });
 
@@ -94,7 +94,6 @@
         $('#select-item a').on('o:resource-selected', function (e) {
             var value = $('.value.selecting-resource');
             var valueObj = $('.resource-details').data('resource-values');
-
             $(document).trigger('o:prepare-value', ['resource', value, valueObj]);
             Omeka.closeSidebar($('#select-resource'));
         });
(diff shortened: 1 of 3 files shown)


application/asset/js/resource-form.js
```diff
@@ -246,6 +246,7 @@
         }
         var value = $('.value.template[data-data-type="' + dataType + '"]').clone(true);
         value.removeClass('template');
+        value.attr('data-term', term);
```
zerocrates, line 249: Is there a reason some of these are using `.attr('data-term' ...` instead of `.data('term', ...` ?

Daniel-KM: I didn't checked all the cases, but in https://github.com/omeka/omeka-s/blob/develop/application/asset/js/resource-form.js#L340, you indicated that there may be some issues without attr. Here, it is useless, so data can be used. But the data is required (else the module cannot determine the property of the value to be created).

application/asset/js/resource-form.js
```diff
@@ -227,16 +226,25 @@
         return values;
     };
 
+    var makeDefaultValue = function (term, dataType) {
+        return makeNewValue(term, dataType)
+            .addClass('default-value')
+            // FIXME When the template is removed clicking directly on the cross, this event is skipped, so the content is removed.
```
zerocrates, line 232: Instead of FIXMEs and TODOs within the code, please file an issue or call attention to an existing issue.

As for this specifically, I can't confirm an issue with the `change` event here, it seems to work as expected. Unless I'm misunderstanding the comment.

application/asset/js/resource-form.js
zerocrates: Okay, you're referring to an existing comment, I see.

I believe that what that comment refers to is other parts of the code using attribute selectors to find the elements: when that functionality is being used, then we need to use `attr`, or else there's no attribute to select against.

application/asset/js/resource-form.js
Daniel-KM: The issue is probably outdated, I'll remove the comment.

application/asset/js/resource-form.js
Daniel-KM: Ok, I convert it into data.

Daniel-KM:
And in fact, this pr is the only one required for the settings improvements; all other code can overridden by a module. This is not like the multiple datatypes improvment.

Daniel-KM:
And in fact, this pr is the only one required for the settings improvements; all other code can be overridden by a module. This is not like the multiple datatypes improvment.

application/asset/js/resource-form.js
```diff
@@ -356,163 +365,189 @@
...
+                    console.log('Failed loading resource template from API');
+                })
+                .always(finalize);
+        } else {
+            finalize();
+        }
+
+        function finalize() {
+            // Remove empty properties, except the templates ones.
+            // TODO Keep properties selected by user (remove the "default-value"?).
```
zerocrates, line 493: I don't think we want to have in the feature to remove empty properties if it doesn't handle keeping the user-selected ones.

Daniel-KM:
I added a commit to add a wrapper to manage sidebars of item set and media like the item one, else it's not possible to determine simply the type of resource to search in the sidebar.

application/asset/js/resource-form.js
Daniel-KM: This is fixed below.

Daniel-KM:

PR #1628, opened 2020-09-07
Overridden chosen-select default css with style.css.

Daniel-KM:
In some cases, the style are overridden by the default css of chosen-select, so this fix allows to keep defined styles.


zerocrates:
Are you actually seeing precedence problems with this? It looks to me like the selectors we have already are parallel with the ones from Chosen, so ours should already be taking over.

Is there a visual issue you can point to that's currently happening here?

PR #1630, opened 2020-09-07
Fixed svg media types for assets.

Daniel-KM:
This is a request that was integrated in 2.1.2 (#1583), but not in the dev version. It may depend on a future merge anyway.


zerocrates:
Thanks for pointing this out.

I've merged master back into develop so the original commit should be on develop now (plus one of your other PRs).

PR #1634, opened 2020-09-17
Multiple data types by property in form

Daniel-KM:
This is the single feature to manage multiple data types in resource template (see #1614).

### DIFF
--- a/application/data/migrations/20200831000000_MutlipleDataTypes.php
+++ b/application/data/migrations/20200831000000_MutlipleDataTypes.php
@@ -1,0 +1,30 @@
+<?php
+namespace Omeka\Db\Migrations;
+
+use Doctrine\DBAL\Connection;
+use Omeka\Db\Migration\MigrationInterface;
+
+class MutlipleDataTypes implements MigrationInterface
+{
+    public function up(Connection $conn)
+    {
+        $sql = <<<SQL
+ALTER TABLE `resource_template_property` CHANGE `data_type` `data_type` LONGTEXT COLLATE utf8mb4_unicode_ci DEFAULT NULL COMMENT '(DC2Type:json_array)';
+SQL;
+        $conn->exec($sql);
+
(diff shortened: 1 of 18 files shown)


application/data/migrations/20200831000000_MutlipleDataTypes.php
```diff
@@ -0,0 +1,30 @@
...
+        $sql = <<<SQL
+UPDATE `resource_template_property`
+SET `data_type` = NULL
+WHERE `data_type` IS NULL OR TRIM(`data_type`) = "";
+SQL;
+        $conn->exec($sql);
+
+        $sql = <<<SQL
+UPDATE `resource_template_property`
+SET `data_type` = CONCAT('["', REPLACE(TRIM(`data_type`), "\n", '","'), '"]')
```
zerocrates, line 25: Quick question here: the handling of newlines and multiple values, that's accounting for a previous version of this that stored the multiple values that way?

I'd imagine we could leave that, and therefore the whole `REPLACE` out, yeah? And for that matter the check for `[` in the WHERE?

Anybody who's been on an intermediate version of one of these could just throw this migration id in the migrations table and skip it.

zerocrates:
I believe the initial version of this had changes to import/export, which are now gone. Is that being separated out or what? It's fine to be in this PR along with the rest of what's here.

If I've just caught you in between changes, that's fine also.

application/data/migrations/20200831000000_MutlipleDataTypes.php
Daniel-KM: Yes, it is related to an old version. I removed it.

Daniel-KM:
I saw that there was the import view that was remaining, but not the controller part. So I rebased and now there is the whole import/export part.

zerocrates:
Thanks for bearing with us on all those change requests.

PR #1643, opened 2020-10-01
Resource value input styles

kimisgold:
This sets nicer container styles within the resource form for value data types outside of the core, while also hiding sorting handles when there is only one value within a property field. It has been tested with the Numeric Data Types and Data Type Geometry modules.


zerocrates:
Issues:

- growing a textarea to be taller stretches out the spacing between the trash and private icons, I believe this is a regression
- shrinking a textarea now makes it align to the middle, the default white background is visible and the borders of the textarea overlap the language chooser, which stays top-aligned.

Perhaps just make the textarea min height larger so it matches the default height and the height of the icons area: there's really no point in letting it shrink smaller than that since it saves no space anyway.

On the icons, `flex-direction: column` instead of `flex-wrap: wrap` would work most easily, I would think.

A question: what's the added "flexiness" of the input-body being used for? Just to center the shorter contents?

kimisgold:
> A question: what's the added "flexiness" of the input-body being used for? Just to center the shorter contents?

Yes. Definitely open to alternative approaches, as I can see there being markup for which this creates problems.

PR #1648, opened 2020-10-13
browse preview block content control

matthewjmckinley:


application/src/Site/BlockLayout/BrowsePreview.php
```diff
@@ -58,6 +59,28 @@ public function form(PhpRenderer $view, SiteRepresentation $site,
...
+            'options' => [
+                'label' => 'Components', // @translate
+                'info' => 'Components to display for each resource. If not set in Site Settings, Heading defaults to resource Title and Body to resource Description', // @translate
+                'value_options' => [
+                    [
+                        'value' => 'resource-heading',
+                        'label' => 'Heading', // @translate
+                    ],
+                    [
+                        'value' => 'resource-body', // @translate
```
zerocrates, line 74: The translate comment here is accidentally on the wrong string

application/src/Site/BlockLayout/BrowsePreview.php
```diff
@@ -23,6 +23,7 @@ public function form(PhpRenderer $view, SiteRepresentation $site,
             'query' => '',
             'heading' => '',
             'limit' => 12,
+            'components' => ["resource-heading","resource-body","thumbnail"],
```
zerocrates, line 26: Just simple style matters: simple strings should be using single quotes, and there should be a space after every comma. Same applies to this same array appearing in `render`.

application/view/common/block-layout/browse-preview.phtml
```diff
@@ -16,9 +16,15 @@ foreach ($this->resources as $resource):
     $body = $bodyTerm ? $resource->value($bodyTerm) : $resource->displayDescription();
 ?>
     <li class="<?php echo $this->resourceType; ?> resource">
-        <?php echo $resource->linkRaw($this->thumbnail($resource, 'medium', ['title' => $heading])); ?>
-        <h4><?php echo $resource->link($heading); ?></h4>
-        <?php if ($body): ?>
+        <?php
```
zerocrates, line 19: My only substantive note: it's not ideal to be checking for the presence of the components in the array here, since it's inside the for loop, so we'll be re-searching over the array for every displayed resource.

These should probably be just checked once at the top and assigned to variables that are actually used in the `if` statements here.

application/view/common/block-layout/browse-preview.phtml
```diff
@@ -16,9 +16,15 @@ foreach ($this->resources as $resource):
...
+        if (in_array('thumbnail', $this->components)):
+            echo $resource->linkRaw($this->thumbnail($resource, 'medium', ['title' => $heading]));
+        endif;?>
+        <h4><?php
```
zerocrates, line 23: This `if` should be moved: as is it will output an empty `h4` element if the heading is disabled, when the `h4` should be removed completely in that case.

PR #1649, opened 2020-10-13
Add optional div wrapper to HTML block

matthewjmckinley:


application/src/Site/BlockLayout/Html.php
```diff
@@ -38,17 +39,38 @@ public function onHydrate(SitePageBlock $block, ErrorStore $errorStore)
...
+        $form->add($div_class);
+
+        return $view->formCollection($form);
     }
 
     public function render(PhpRenderer $view, SitePageBlockRepresentation $block)
     {
-        return $block->dataValue('html', '');
+        $html_block = $block->dataValue('html', '');
+        $div = htmlspecialchars($block->dataValue('divclass'));
```
zerocrates, line 63: We conventionally use the framework's `escapeHtml` helper for this kind of escaping.

I think `$divClass` is probably a better variable name here, too. `$div` is a little confusing, maybe.

application/src/Site/BlockLayout/Html.php
```diff
@@ -38,17 +39,38 @@ public function onHydrate(SitePageBlock $block, ErrorStore $errorStore)
     public function form(PhpRenderer $view, SiteRepresentation $site,
         SitePageRepresentation $page = null, SitePageBlockRepresentation $block = null
     ) {
-        $textarea = new Textarea("o:block[__blockIndex__][o:data][html]");
-        $textarea->setAttribute('class', 'block-html full wysiwyg');
+        $form = new Form();
+        $html = new Element\Textarea("o:block[__blockIndex__][o:data][html]");
+        $html->setAttribute('class', 'block-html full wysiwyg');
+        $div_class = new Element\Text("o:block[__blockIndex__][o:data][divclass]");
```
zerocrates, line 45: Our style would be to use camelCase for variables

PR #1661, opened 2020-11-03
Added a mode "auto" to cli to use proc_open() when available.

Daniel-KM:
This mode avoids the admin to have to update the config. `proc_open()` is more useful, since it log errors, so it is set as default when available.

### DIFF
--- a/application/config/module.config.php
+++ b/application/config/module.config.php
@@ -195,7 +195,7 @@ return [
         'sslcafile' => null,
     ],
     'cli' => [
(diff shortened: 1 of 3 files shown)


zerocrates:
I like this idea.

Even though it's minimal processing, I'd prefer to see the disabled function ini setting parsed only once, instead of being done for each call to check a function.

Daniel-KM:
Ok, I simplified it.

zerocrates:
That's great, thanks.

PR #1666, opened 2020-11-20
List of Pages Site Block

matthewjmckinley:


zerocrates:
So, a question here: how much of the JSTree code here is identical to the code used in the nav form? It looks like a pretty significant amount of it is, so I'm wondering if we shouldn't be pulling that out so it's not so duplicated.

Especially as it appears that the "plugins" can get chosen right at the time of initializing the tree, it seems like we can probably just have them defined somewhere in common and loaded both places.

matthewjmckinley:
You're right, there is a lot of replication and I knew when committing it was not ideal. The whole purpose of this was to remove the editLink button from each site in the List of Pages block, since being able to edit and save site settings from within the block added needless complexity and could introduce problems. 

I remember trying to use the existing jstree code in site-navigation.js, and getting it to load correctly, but no matter how I tried I could not get the editLink link/icon to disappear from each page 'row', even when it was explicitly NOT included in the plugin list when the jstree was initialized.

Perhaps there is a method I missed, but the only way I could to get the editLink to truly disappear was to copy over the whole .jstree setup from site-navigation.js and remove the editLink portion.

zerocrates:
OK, I think just two more things:

- I think the JS additions for this block should be in their own JS file rather than being embedded within the existing site-page JS
- I'm a little concerned about the str_repeat strategy in use for initially rendering the tree... the resulting structure's not really a tree

PR #1667, opened 2020-11-23
Allowed to search by id with api endpoint.

Daniel-KM:
In internal api, it is possible now to search with a list of ids. It is not possible by endpoint, so this fixes it.

### DIFF
--- a/application/src/Controller/ApiController.php
+++ b/application/src/Controller/ApiController.php
@@ -229,18 +229,16 @@ class ApiController extends AbstractRestfulController
     /**
      * Check if the request contains an identifier.
      *
-     * This method overrides parent in order to allow to query on ids.
+     * This method overrides parent in order to allow to query on one or
+     * multiple ids.
      *
      * {@inheritDoc}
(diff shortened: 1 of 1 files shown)


zerocrates:
I'm wondering if we might just want to turn off completely the ability for this to look up "the ID" from the query string, since that's really not the way our REST API is intended to work. So to instead override this to just look only at the RouteMatch for an ID.

Daniel-KM:
It is very useful to request the api with a list of ids, so if `id=xxx` is useless, `id[]=xxx` should be kept. Is it that you mean?

zerocrates:
I think we're on the same page with your revised commit here: I just meant to use the RouteMatch only in this method so that a user can query on either `?id=X` or `?id[]=X&id[]=Y` and we'll still correctly do a "search" instead of the current behavior where the Laminas layer is seeing an `id` key and interpreting the request as a "read."

Daniel-KM:
Yes, so my second commit is good: only `/api/items/xxx` is a read, everything else is a search.

zerocrates:
OK, great.

PR #1670, opened 2020-11-25
Support text/json+oembed (Soundclound)

Avnerus:
Hi, 
I tried to embed a _Soundcloud_ link and the the oEmbed directive was not found.
I noticed that they use the type`text/json+oembed` instead of `application/json+oembed`.
After applying the following fix, Soundcloud embedding works.

/Avner

### DIFF
--- a/application/src/Media/Ingester/OEmbed.php
+++ b/application/src/Media/Ingester/OEmbed.php
@@ -77,7 +77,7 @@ class OEmbed implements IngesterInterface
 
         $document = $response->getBody();
         $dom = new Query($document);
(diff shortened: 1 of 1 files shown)


zerocrates:
I assume there's no chance of Soundcloud changing things on their end...

It's a little annoying to be dealing with a very short spec like oEmbed that specifically says

> The `type` attribute must contain either `application/json+oembed` for JSON responses, or `text/xml+oembed` for XML.

but to encounter something like this anyway.

zerocrates:
On the other hand we already accept link rel "alternative" here so I suppose there's no need to stand on strict correctness and I can't see it harming anything, so I merged it.

Thanks for your contribution.

PR #1673, opened 2020-12-09
Update el_GR.po

kolydart:
fix typo in translation


zerocrates:
Hi!

We manage the translation strings through [Transifex](https://www.transifex.com/omeka/omeka-s/), so any changes and fixes should go through there. If you've made the change in Transifex already, we'll pick it up at the next release. If not, you can sign up and join at the link I posted.

zerocrates:
I've applied this manually to the Transifex Greek strings.

PR #1683, opened 2021-01-27
Feature/vips

Daniel-KM:
A new thumbnailer, that has the main quality to be very quick (5 to 10 times quicker than ImageMagick, as the [ad says](https://github.com/libvips/libvips/wiki/Speed-and-memory-use)). It is the one [used by wikipedia](https://www.mediawiki.org/wiki/Extension:VipsScaler), but it is implemented directly here, without extension. 

I added a second thumbnailer "auto", in order to simplify the choice for the admin. 

A point to check is that it is recommended to use a recent version, because there is an interesting feature, the possibility to crop the square thumbnail according to the point of attention, that may not be the center (gravity). Nevertheless, it is not yet in the stable packages of the common linux server distributions.

As usual, it is already implemented in a module, the module [Image Server](https://gitlab.com/Daniel-KM/Omeka-S-module-ImageServer), since some weeks. Indeed, it can be used to tile jpeg2000 and pyramidal tiff dynamically without the need to pre-tile images. Last version can manage pyramidal iiif images too.


zerocrates:
I'll have to get a hold of vips so I can try this out.

In principle I'm in favor of it, nothing wrong with more options.

application/src/File/Thumbnailer/Vips.php
```diff
@@ -0,0 +1,229 @@
...
+                'high',
+                'attention',
+                'entropy',
+                // "all" does not crop as square.
+                // 'all',
+            ];
+            $mapImagickToVips = [
+                'northwest' => 'high',
+                'north' => 'high',
+                'northearth' => 'high',
```
zerocrates, line 111: Looks like a typo for "northeast" here?

application/src/File/Thumbnailer/Vips.php
```diff
@@ -0,0 +1,229 @@
...
+            $crop = '';
+        }
+
+        $tempFile = $this->tempFileFactory->build();
+        $tempPath = $tempFile->getTempPath() . '.jpg';
+        $tempPathCommand = $tempPath . '[background=255 255 255,optimize-coding]';
+        $tempFile->delete();
+
+        $command = sprintf(
+            '%s thumbnail %s %s %d --height %d%s%s --size both --linear --intent absolute',
```
zerocrates, line 136: From my understanding, "both" isn't appropriate for "size" here: the other thumbnailers don't scale up, only down (except for the "square" type).

zerocrates:
Can you explain what's going on with the usage of vips vs vipsthumbnail? It seems like vipsthumbnail has support for doing pretty much everything we'd want: specifying the size, scale up/down/both, crop type with `--smartcrop`, `--rotate`, the options for output...

It seems to me like this could just always use `vipsthumbnail`, so I'm wondering what the rationale is for the other strategy.

application/src/File/Thumbnailer/Vips.php
Daniel-KM: This is fixed in last commit

application/src/File/Thumbnailer/Vips.php
Daniel-KM: This is fixed in last commit.

Daniel-KM:
About the choice on "vips"/"vipsthumbnail" : i created the thumbnailer for the module Image Server and it requires some complex parameters to manage iiif requirements, so i worked with vips and the same here. Options are not exactly the same, so it's simpler to keep the same command line whatever the help said.

Daniel-KM:
I integrated it in a separated module [vips](https://github.com/Daniel-KM/Omeka-S-module-Vips) and it will be removed from module [Image Server](https://github.com/Daniel-KM/Omeka-S-module-ImageServer).

PR #1684, opened 2021-01-28
Search type and language case insensitively (fix #1672).

### LINKED ISSUE #1672: Lang tags should be compared case insensitively
BCP 47 2.1.1 calls for language tags to be treated as case insensitive at all times.

We get some case insensitivity for free from our MySQL collation, but in places where we're doing comparisons in PHP code, we're currently case-sensitive. I'm thinking particularly of the "lang" option for `value()` in representations

Daniel-KM:


zerocrates:
Looks good, thanks.

PR #1485, opened 2019-12-30
Added option to specify the label information per site.

Daniel-KM:

PR #1689, opened 2021-02-11
Added the information about property in site settings.

Daniel-KM:
Currently, the only admin setting is used inside sites.


zerocrates:
This is essentially the same as the earlier, now merged, #1485, so I'll close this one.

PR #1692, opened 2021-03-01
Site localization: display item and item sets titles and values based on site locale

ManOnDaMoon:
This was originally submitted as a omeka-s-themes/foundation-s PR: https://github.com/omeka-s-themes/foundation-s/pull/25

Although the scope of functionality is similar, the implementation has been redone and a lot more is handled within the core PHP of Omeka S and not in views anymore.

### Description

This is a new feature submission that enables site localization by:

* hiding or showing values based on their language IETF tag
* retrieving resources' title and body in the adequate language based on site settings or resource template

Use case: Items store several language values for a same property and users want to create language specific sites.

### Features:

* Display items, property values, item sets, linked resources in the site language as per the site settings

Site "Language" settings have been grouped to introduce the following 2 settings:
* New site setting "Filter values based on site localization": global option to enable the feature. When checked, resources' property values will be shown only if their associated language IETF tag matches the site locale setting. Values without specified language will be shown in all cases, especially when language is not applicable (e.g.: serial numbers or IDs, etc.).
* New site setting "Show values locale labels": this conditionally shows or hides the language label (useful when the previous setting is enabled and thus the label should be unique across the site).

Additionnaly, a "Site Switcher" navigation link is introduced that can be used as a toggle between languages.

You can have a look at the feature live on https://collection.manondamoon.com/s/en - a flag icon in the top bar is used to switch between sites with different languages. If more languages are to be supported, it is only a matter of adding additional navigation links to the desired locale.

Let me know if this feature is of interest.

Have a nice day

Laurent


application/config/module.config.php
```diff
@@ -529,6 +529,7 @@
             'url' => Site\Navigation\Link\Url::class,
             'browse' => Site\Navigation\Link\Browse::class,
             'browseItemSets' => Site\Navigation\Link\BrowseItemSets::class,
+            'siteSwitcher' => Site\Navigation\Link\SiteSwitcher::class,
```
zerocrates, line 532: Could you separate out the siteSwitcher stuff into a separate branch/PR? I can see how they kind of work together but for review purposes, I think we'd like it separate from the values feature.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -501,11 +501,28 @@ public function displaySubjectValues($page = null, $perPage = null, $property =
...
+        $template = $this->resourceTemplate();
+        if ($template && $template->titleProperty()) {
+            $titleTerm = $template->titleProperty()->term();
+        } else {
+            $titleTerm = 'dcterms:title';
+        }
+
+        if ($lang !== null) {
+            if ($titleValue = $this->value($titleTerm, ['lang' => $lang])) {
+                $title = $titleValue->value();
```
zerocrates, line 518: In these spots where you're getting a value to return, you should be casting the value object to string: that uses the toString method of the value's data type, which isn't necessarily just returning the `->value()`.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -537,7 +555,15 @@ public function displayDescription($default = null)
             $descriptionTerm = 'dcterms:description';
         }
 
-        $description = (string) $this->value($descriptionTerm, ['default' => $default]);
+        if ($lang !== null) {
+            if ($descriptionValue = $this->value($descriptionTerm, ['default' => $default, 'lang' => $lang])) {
+                $description = $descriptionValue->value();
```
zerocrates, line 560: Same comment about casting vs `->value()` here.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -501,11 +501,28 @@ public function displaySubjectValues($page = null, $perPage = null, $property =
...
+            $titleTerm = 'dcterms:title';
+        }
+
+        if ($lang !== null) {
+            if ($titleValue = $this->value($titleTerm, ['lang' => $lang])) {
+                $title = $titleValue->value();
+            }
+        }
+
+        if (!isset($title) || $title === null) {
```
zerocrates, line 522: I'd prefer not to have these "is not set or is null" checks: it's cleaner probably to just define `$title` as null near the top and then not need the isset check.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -537,7 +555,15 @@ public function displayDescription($default = null)
...
         }
 
-        $description = (string) $this->value($descriptionTerm, ['default' => $default]);
+        if ($lang !== null) {
+            if ($descriptionValue = $this->value($descriptionTerm, ['default' => $default, 'lang' => $lang])) {
+                $description = $descriptionValue->value();
+            }
+        }
+
+        if (!isset($description) || $description === null) {
```
zerocrates, line 564: Same comment about `!isset` here.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -568,6 +594,7 @@ public function displayResourceClassLabel($default = null)
      * @return string
      */
     public function linkPretty(
+        $lang = null,
```
zerocrates, line 597: Adding `lang` as the first parameter to linkPretty messes with the signature which we probably don't want to do.

application/src/Api/Representation/MediaRepresentation.php
```diff
@@ -265,9 +265,10 @@ public function item()
      * Change the fallback title to be the media's source, if it exists.
      *
      * @param string|null $default
+     * @param string|null $lang
      * @return string|null
      */
-    public function displayTitle($default = null)
+    public function displayTitle($default = null, $lang = null)
```
zerocrates, line 271: You should be passing this `$lang` here to the parent method call for it to have any effect.

application/src/Form/SiteSettingsForm.php
```diff
@@ -124,6 +134,32 @@ public function init()
             ],
         ]);
 
+        $langFieldset->add([
+            'name' => 'filter_locale_values',
+            'type' => 'checkbox',
+            'options' => [
+                'label' => 'Filter values based on site localization', // @translate
```
zerocrates, line 141: "Localization" is likely to get confusing here with a Language section and Locale setting, it should probably say "locale" here.

application/view/common/resource-values.phtml
```diff
@@ -32,12 +37,14 @@ $labelInfo = $this->setting('property_label_information');
                 $class[] = 'private';
             }
             ?>
-            <div class="<?php echo implode(' ', $class); ?>" lang="<?php echo $escape($value->lang()); ?>">
-                <?php if ($language = $value->lang()): ?>
-                <span class="language"><?php echo $language; ?></span>
+            <?php if (($filterLocale == 0) || ($valueLang == '') || ($valueLang == $lang)): ?>
```
zerocrates, line 40: I'd have expected the actual filtering of values to happen at the resource representation's `displayValues` method. Partially just because we already have _other_ "lang-filtering" functionality there in `value()`.

If it's going to be here (or anywhere, I guess), then we have to handle some things that the `value()` implementation already does, like comparing langs case-insensitively.

application/view/common/resource-values.phtml
```diff
@@ -32,12 +37,14 @@ $labelInfo = $this->setting('property_label_information');
...
+            <div class="<?php echo implode(' ', $class); ?>" lang="<?php echo $escape($valueLang); ?>">
+                <?php if (($showLocale == 1) && ($valueLang)): ?>
+                <span class="language"><?php echo $valueLang; ?></span>
                 <?php endif; ?>
-                <?php echo $value->asHtml(); ?>
+                <?php echo $value->asHtml($lang); ?>
```
zerocrates, line 45: This is one of the only places (maybe _the_ only place?) where the `$lang` setting gets used regardless of if "filterLocale" is enabled. With it _disabled_ it seems inconsistent that we'd modify the displayed titles of linked resources when listing values (the practical effect here), but not when browsing, etc.

application/view/common/advanced-search/item-sets.phtml
```diff
@@ -30,6 +33,7 @@ if (!$ids) {
                 ],
                 'options' => [
                     'disable_group_by_owner' => $this->status()->isSiteRequest(),
+                    'lang' => ($filterLocale == 1 ? $lang : null),
```
zerocrates, line 36: Just for simplicity/cleanliness it's probably best to just cast `filterLocale` to bool when getting it at the top of each of these files where it's used and thereby avoid all the `== 0`/`== 1` comparisons.

application/src/Form/SiteSettingsForm.php
```diff
@@ -124,6 +134,32 @@ public function init()
...
+                'info' => 'Check to show only values matching the site language setting and values without locale ID.',
+            ],
+            'attributes' => [
+                'id' => 'filter_locale_values',
+                'value' => (bool) $settings->get('filter_locale_values', false),
+            ],
+        ]);
+
+        $langFieldset->add([
```
zerocrates, line 150: This feature should _also_ probably be its own thing (for one it's much simpler and therefore easier to merge) but I think we can probably deal with them together.

We could also merge it separately, though.

zerocrates (commented):
Thanks for making this conversion of your initial code. This definitely makes more sense to be in the core than an individual theme, I think.

I like how this works, I think in general it's a pretty workable layer on top of the existing systems that's mostly just providing a more convenient UI/setting access to features we already have but that currently require custom theme support. Having this generally accessible would be a very nice improvement.

I do have some comments/suggestions/requested changes, which should be included in this review.

application/config/module.config.php
ManOnDaMoon: OK, SiteSwitcher removed in f72b394
I'll submit a separate PR.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
ManOnDaMoon: OK updated in 81ec357

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
ManOnDaMoon: OK updated in 81ec357

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
ManOnDaMoon: OK updated in 81ec357

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
ManOnDaMoon: 81ec357

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
ManOnDaMoon: You're totally right. Updated in 81ec357

application/src/Form/SiteSettingsForm.php
ManOnDaMoon: OK updated in cccdf45

application/view/common/advanced-search/item-sets.phtml
ManOnDaMoon: OK updated in 5c6f968

application/src/Api/Representation/MediaRepresentation.php
ManOnDaMoon: OK fixed in 52200f7

application/view/common/resource-values.phtml
ManOnDaMoon: You're right we should be able to filter this when retrieving the values. I don't see immediately if `AbstractResourceEntityRepresentation::displayValues()` uses exactly the same implementation as `AbstractResourceEntityRepresentation::value()` for each value, but I'll have a closer look.

application/view/common/resource-values.phtml
ManOnDaMoon: Took me some time to figure this out, nice catch. 69e554e

application/src/Form/SiteSettingsForm.php
ManOnDaMoon: Once you filter the values based on the language, you end up with item show pages cluttered with now kind of useless labels, a lot of them. That's my experience on this but if your really prefer to separate this, please tell me.

zerocrates: I think It'll be better as a separate PR, though I do agree it becomes more important to have with the "filter by locale" feature turned on.

Among other reasons, It'll be easier for us to track it as a separate new feature in release notes and things like that.

A side note: even though it's only checked once, I think the same "cast to bool" advice applies with this check as well.

application/view/common/resource-values.phtml
ManOnDaMoon: ed9336a introduces this filtering being done within `AbstractResourceEntityRepresentation` display methods.

application/src/Form/SiteSettingsForm.php
ManOnDaMoon: OK removed in 1acfc9f

ManOnDaMoon:
I've reverted manually the changes. There are some use cases I'm unable to test without rebasing myself, as I need the latest changes from develop. Usually rebasing to develop breaks my working environment, I'll have a go and let you know.

ManOnDaMoon:
I rebased the branch on develop with the latest changes from yesterday that created conflicts in some of the views.
I forced reverted the changes to `AbstractResourceEntityRepresentation`, thought you should know in case you checked it out locally.
The lastest version of `AbstractResourceEntityRepresentation` made it a little different to handle the `displayTitle` and `displayDescription` part and this rebase was useful to update the feature implementation.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -526,18 +544,32 @@ public function displayTitle($default = null)
...
         if ($template && $template->descriptionProperty()) {
             $descriptionTerm = $template->descriptionProperty()->term();
         } else {
             $descriptionTerm = 'dcterms:description';
         }
 
-        $description = (string) $this->value($descriptionTerm, ['default' => $default]);
+        if ($lang !== null) {
+            if ($descriptionValue = $this->value($descriptionTerm, ['default' => $default, 'lang' => $lang])) {
+                if (!empty($descriptionValue->valueResource())) {
```
zerocrates, line 562: This is a new change, right?

The prior code would just print URIs for "descriptions" that were Resource type... (since that's what the toString for a resource does). This PR should maintain that same behavior: even if we wanted to change it, we'd want to change it for both the "filter locale" and "non filter" paths simultaneously.

You pretty much don't need any of this extra logic in that case.

ManOnDaMoon: Actually before this commit. this was not printing URIs nor anything,the description ended up blank.  Which seemed a little curious to me. Let me check this back, I’m wondering if this is something to do with my « values resources » being CustomVocab data types in my test data.

zerocrates: If it's a problem specifically with CustomVocab "resource" values, that probably indicates an issue with CustomVocab rather than with this code.

ManOnDaMoon:
On a side note, you've mentioned the possibility of introducing a `valuesFiltered()` method earlier and the comment got lost during the rebase.
I had a go with this on a separate branch that can be seen there if you want to have a look: https://github.com/ManOnDaMoon/omeka-s/commit/d2be682e4a16a76402a28e34e46c580e3e795f3a

zerocrates:
I've done a squash/rebase of _most_ of this: most significantly I've excluded your "edge case" description commit. I think that needs further discussion before it could go in, and it may not need to at all.

Thanks for all your work on this. I'll be closing this PR, but additional discussion about changes/improvements could happen on new PRs or issues, that's fine.

zerocrates:
Your branch doing "filteredValues" is interesting, it's just about what I was thinking of, yeah.

Though the real "plus" of it would be to make it so it works (and works efficiently) so it can be used by `value()` also. I think yours can pretty much do this already by using the `term` filter but actually making the substitution is sort of the key to that idea. (`value` then has to sort of redundantly throw away the extra info, the property and so on but... it's probably not worth it to have a mode that just directly returns an array of values for a single term just to avoid that)

PR #1712, opened 2021-04-29
Site list thumbnail

matthewjmckinley:
Can you please review Kim? I did some minimal CSS styling on list-of-site and site-list-entry to make it legible, but didn't want to fine-tune without your input. Happy to change anything if needed!


application/view/common/site-list-entry.phtml
```diff
@@ -1,7 +1,11 @@
 <?php
+$thumbnail = $showThumbnail ? $site->thumbnail() : null;
 $summary = $showSummary ? $site->summary() : null;
 ?>
 <div class="site">
+    <?php if ($thumbnail): ?>
+    <p class="site-thumbnail"><?php echo $this->thumbnail($thumbnail, 'square', ['class' => 'site-thumbnail-image']); ?></p>
```
kimisgold, line 7: A nitpick, but a `div` wrapper would be more appropriate as a wrapper for an image.

application/asset/sass/_screen.scss
```diff
@@ -4111,6 +4112,11 @@ body.minimal {
             margin: 0 0 0 $spacing-medium;
             line-height: 20px;
         }
+        
+        .site-thumbnail-image {
+            max-width: 25%;
+            float: left;
+        }
```
kimisgold, line 4119: Where we're floating the thumbnails, I think a right margin (maybe 1rem?) on the thumbnails would be a good default.

application/asset/sass/page-blocks.scss
```diff
@@ -222,5 +223,11 @@
     .site-summary {
         margin: 0 0 0 .5rem;
```
kimisgold, line 224: Maybe condense the margin styles here. I don't think we even need the left margin on the summary.

kimisgold (commented):
Just a few details, but otherwise looks great to me.

PR #1714, opened 2021-05-05
Allowed to authenticate via local session for api requests (fix #1613).

### LINKED ISSUE #1613: How to access the api endpoint in admin?
Some dynamic features (like [Item copy](https://omeka.org/s/modules/ItemCopy/)) requires access to the api endpoint in admin board, but this one is only available when a api token is set, so it's complex to use for normal users. 

Is there a way to use the standard authentification in the api too? It will allow to improve the admin board simpler. Of course, it's easy to create a quick module to override or to proxify the endpoint, so the main question is it an interesting feature or not for the core?

Daniel-KM:
This is a simpler way to fix #1613. 
it's included in module [Guest](https://gitlab.com/Daniel-KM/Omeka-S-module-Guest) for now, but probably simpler to manage in core, because it is useful in many current (sidebar, search autocompletion, module [item copy](https://omeka.org/s/modules/ItemCopy/)…) and future cases (ajaxification of the admin user interface).

### DIFF
--- a/application/src/Mvc/MvcListeners.php
+++ b/application/src/Mvc/MvcListeners.php
@@ -263,8 +263,9 @@ class MvcListeners extends AbstractListenerAggregate
             return;
         }
 
-        $identity = $event->getRequest()->getQuery('key_identity');
-        $credential = $event->getRequest()->getQuery('key_credential');
+        $request = $event->getRequest();
+        $identity = $request->getQuery('key_identity');
+        $credential = $request->getQuery('key_credential');
 
         if (is_null($identity) || is_null($credential)) {
             // No identity/credential key to authenticate against.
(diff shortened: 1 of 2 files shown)


zerocrates:
I think we're better off keeping the regular api endpoint as something that doesn't look at the session ever. Feels like something that will complicate testing and so on, and also probably more importantly there's the security implications of allowing CORS, and even same-origin requests to the API (this is safe if the API never uses the session, has CSRF implications if it doesnt). Normal POSTs, etc. in the admin UI use CSRF tokens, but the API does not and that safety depends on the cookie/session not being used there.

I don't think it's a problem to have the possibility to use the API based on the session cookie, but it would need to come with CSRF-type protection at least, I think, which probably rules out this kind of simplest solution.

PR #1716, opened 2021-05-05
Improved the link to browse resources in user bar.

Daniel-KM:
The link to the items should be limited to the current site items.

Another issue: when the site title is long, the user bar is not pretty, but @kimisgold will fix it better than me.


zerocrates:
Can you make an issue for the title length thing? We'll lose track of it when we close this PR, probably.

PR #1726, opened 2021-06-08
Show Locale Label setting

ManOnDaMoon:
New site setting "Show values locale labels": this conditionally shows or hides the language label on item-sets browse and item show views.
Useful when combined with #1692.


zerocrates:
I do anticipate a minor problem here: because it's defaulting to false, this will turn off the labels for all existing sites upon upgrading.

I think you need to switch that default to true, or else add a migration to set true options for all existing sites.

Also I think on the label for the setting let's go with "Show locale labels for values"

zerocrates:
Great, thanks.

PR #1731, opened 2021-06-23
Asset block

kimisgold:
This site page block allows the user to display multiple assets. Each asset can link to a page within the site, provide an alternate label for that link, and also appear alongside a short caption.

The assets block can be used to display assets on a page that don't exist as item or media within the site resource pool, like for user guides or about pages. This is also a possible means to more easily build a customized home/landing page. With this in mind, the block includes the ability to set custom classes that theme designers can use to more easily style and position. The block also offers some basic pre-styled layout options, similar to those offered in the media block: float left, float right, and centered.


application/src/Site/BlockLayout/Asset.php
```diff
@@ -0,0 +1,92 @@
...
+    
+    public function onHydrate(SitePageBlock $block, ErrorStore $errorStore)
+    {
+        $data = $block->getData();
+        $block->setData($data);
+      }
+      
+    public function alignmentClassSelect(PhpRenderer $view, SitePageBlockRepresentation $block = null
+    ) {
+        $alignmentLabels = ['default', 'float left', 'float right', 'center'];
```
zerocrates, line 32: These should be individually marked as translatable with  `// @translate`

application/src/Site/BlockLayout/Asset.php
```diff
@@ -0,0 +1,92 @@
...
+        $alignmentValues = ['default', 'left', 'right', 'center'];
+        $alignment = $block ? $block->dataValue('alignment', 'default') : 'default';
+        $select = new Select('o:block[__blockIndex__][o:data][alignment]');
+        $select->setValueOptions(array_combine($alignmentValues, $alignmentLabels))->setValue($alignment);
+        $selectLabel = 'Alignment'; // @translate
+        $select->setAttributes(['title' => $selectLabel, 'aria-label' => $selectLabel]);
+        $html = '<div class="field"><div class="field-meta">';
+        $html .= '<label class="thumbnail-option" for="o:block[__blockIndex__][o:data][alignment]">' . $selectLabel . '</label></div>';
```
zerocrates, line 40: `for` connects to an input using an `id`, not a name, so this doesn't properly associate with the input.

You'll need to set an id for the select (if you do it through the options/attributes system for the form element object, you can just print the label with the form label helper too and it will do the proper `for` for you).

zerocrates: `thumbnail-option` class?

application/view/common/asset.phtml
```diff
@@ -0,0 +1,65 @@
+<?php
```
zerocrates, line 1: just "asset" is probably not the best name for this partial

application/view/common/asset.phtml
```diff
@@ -0,0 +1,65 @@
...
+        <li class="undo">' . $this->hyperlink('', '#', ['class' => 'o-icon-undo button', 'title' => $translate('Restore attachment')]) . '</li>
+    </ul>
+    <input type="hidden" class="asset-page-id asset-option" name="o:block[__blockIndex__][o:data][__attachmentIndex__][page]" data-page-title="" data-page-url="">
+    <input type="hidden" class="asset asset-option" name="o:block[__blockIndex__][o:data][__attachmentIndex__][id]">
+    <input type="hidden" class="alternative-link-title asset-option" name="o:block[__blockIndex__][o:data][__attachmentIndex__][alt_link_title]">
+    <input type="hidden" class="asset-caption asset-option" name="o:block[__blockIndex__][o:data][__attachmentIndex__][caption]">
+    ';
+?>
+<div class="asset-attachments-form" data-site-id="<?php echo $siteId; ?>" data-page-api-url="<?php echo $apiUrl; ?>">
+    <a href="#" class="collapse" aria-label="<?php echo $this->translate('Collapse'); ?>" title="<?php echo $this->translate('Collapse'); ?>"><h4><?php echo $translate('Attachments'); ?></h4></a>
```
zerocrates, line 20: Maybe just "Assets" here as the label? If we want to keep "Attachments" as just what they are currently.

application/view/common/asset.phtml
```diff
@@ -0,0 +1,65 @@
...
+                $caption = $attachment['caption'];
+            ?>
+            <?php $thumbnail = '<div class="thumbnail"><img src="' . $asset->assetUrl() . '" alt="' . $asset->name() . '"></div>'; ?>
+            <div class="asset-title"><?php echo $thumbnail; ?><?php echo $asset->name(); ?></div>
+            <ul class="actions">
+                <li><?php echo $this->hyperlink('', '#', ['class' => 'asset-options-configure o-icon-configure button', 'title' => $translate('Open asset options')]); ?></li>
+                <li class="delete"><?php echo $this->hyperlink('', '#', ['class' => 'o-icon-delete button', 'title' => $translate('Delete attachment')]); ?></li>
+                <li class="undo"><?php echo $this->hyperlink('', '#', ['class' => 'o-icon-undo button', 'title' => $translate('Restore attachment')]); ?></li>
+            </ul>
+            <input type="hidden" class="asset-page-id asset-option" name="o:block[__blockIndex__][o:data][__attachmentIndex__][page]" value="<?php echo $pageId; ?>" data-page-title="<?php echo ($page !== '') ? $page->title() : ''; ?>" data-page-url="<?php echo ($page !== '') ? $page->siteUrl() : ''; ?>">
```
zerocrates, line 41: Data included in these attributes needs to be escaped.

application/view/common/block-layout/asset.phtml
```diff
@@ -0,0 +1,21 @@
+<div class="assets <?php echo $className; ?> <?php echo $alignment; ?>">
+  <?php foreach ($attachments as $attachment): ?>
+    <?php $asset = $attachment['asset']; ?>
+    <div class="asset">
+        <?php $assetImage = '<img src="' . $asset->assetUrl() . '">'; ?>
```
zerocrates, line 5: use `$this->thumbnail` here to get the alt text for the asset included

application/view/common/block-layout/asset.phtml
```diff
@@ -0,0 +1,21 @@
...
+        <?php if (array_key_exists('page', $attachment)): ?>
+            <?php $page = $attachment['page']; ?>
+            <a href="<?php echo $page->siteUrl(); ?>">
+                <?php echo $assetImage; ?>
+                <?php $altLinkTitle = ($attachment['alt_link_title'] !== '') ? $attachment['alt_link_title'] : false; ?>
+                <span class="link-title"><?php echo ($altLinkTitle) ? $altLinkTitle : $page->title(); ?></span>
```
zerocrates, line 11: needs to be escaped

application/view/common/block-layout/asset.phtml
```diff
@@ -0,0 +1,21 @@
...
+            </a>
+        <?php else: ?>
+            <?php echo $assetImage; ?>
+        <?php endif; ?>
+        <?php if ($attachment['caption'] !== ''): ?>
+        <div class="caption"><?php echo $attachment['caption']; ?></div>
```
zerocrates, line 17: needs to be escaped

application/view/common/asset.phtml
```diff
@@ -0,0 +1,65 @@
...
+  <h4>Options</h4>
+</a>
+<div class="collapsible">
+  <div class="field">
+    <div class="field-meta">
+      <label for="o:block[__blockIndex__][o:data][className]"><?php echo $translate('Class'); ?></label>
+    </div>
+    <div class="inputs">
+      <?php $className = (isset($block['className'])) ? $block['className'] : ''; ?>
+      <input type="text" name="o:block[__blockIndex__][o:data][className]" value="<?php echo $className; ?>">
```
zerocrates, line 61: Just always escape or use helpers for values going into inputs like this, basically.

application/view/common/block-layout/asset.phtml
```diff
@@ -0,0 +1,21 @@
+<div class="assets <?php echo $className; ?> <?php echo $alignment; ?>">
```
zerocrates, line 1: escape these

application/view/common/asset.phtml
```diff
@@ -0,0 +1,65 @@
...
+        <button type="button" class="add-asset-attachment" data-sidebar-content-url="<?php echo $this->escapeHtml($this->url('admin/default', ['controller' => 'asset', 'action' => 'sidebar-select'])); ?>"><?php echo $translate('Add asset'); ?></button>
+    </div>
+</div>
+<a class="collapse" href="#" aria-label="<?php echo $translate('collapse'); ?>">
+  <h4>Options</h4>
+</a>
+<div class="collapsible">
+  <div class="field">
+    <div class="field-meta">
+      <label for="o:block[__blockIndex__][o:data][className]"><?php echo $translate('Class'); ?></label>
```
zerocrates, line 57: This _should_ be the same `for` problem as above but isn't seemingly for some reason at runtime... is there a script that is copying the input name to its id? Odd...

application/asset/js/asset-form.js
```diff
@@ -8,6 +8,7 @@
             Omeka.openSidebar(sidebar);
             Omeka.populateSidebarContent(sidebar, $(this).data('sidebar-content-url'));
             selectingForm = $(this).closest('.asset-form-element');
+            $(this).addClass('asset-selecting-button');
```
zerocrates, line 11: Previously discussed, but look at separating this asset-block specific functionality into the asset block's own code rather than having it here in the generic asset form.

zerocrates (changes requested):

application/src/Site/BlockLayout/Asset.php
kimisgold: > `for` connects to an input using an `id`, not a name, so this doesn't properly associate with the input.

I modeled this off the [alignment option used in Media Embed](https://github.com/omeka/omeka-s/blob/develop/application/src/Site/BlockLayout/Media.php#L53). They both seem to be generating an `id`, and associating properly. Am I doing something differently here?

zerocrates: I tried one of these and it wasn't activating the input when you clicked the label... I'll have to double check.

On the general point of the ID being generated... I'll have to check that too... the only one I saw an ID on in the inspector I believe was actually the other input, the one you're hand-writing the markup for. Which is odd.

kimisgold:
I'd love some advice on how to deal with sidebar management. If I have the asset sidebars open and switch gears to managing another block that uses sidebars, the asset sidebars stay open. I'm struggling with coming up with the logic to decide what's open at any given time.

PR #1732, opened 2021-06-23
List of pages 2

matthewjmckinley:
Changes for ListOfPages block to start from empty page list and build from sidebar


application/view/common/block-layout/list-of-pages.phtml
```diff
@@ -4,7 +4,11 @@
             foreach ($pages as $page) {
                 $pageName = $page['text'];
                 $pageLink = $page['data']['url'];
-                echo '<li>' . $this->hyperlink($pageName, $pageLink) . '</li>';
+                if (empty($pageLink)) {
+                    echo '<li>[Missing Page]</li>';
```
zerocrates, line 8: This "[Missing Page]" text needs to be translated.

PR #1756, opened 2021-08-03
Order block layouts alphabetically (fix #1755)

### LINKED ISSUE #1755: Order block layouts alphabetically
Following a discussion about #1751 we've decided to order block layouts alphabetically. See [this comment](https://github.com/omeka/omeka-s/issues/1751#issuecomment-889413550) for implementation ideas.

jimsafley:


zerocrates:
A minor point but: the collator is default case-insensitive (and that's probably what we want), so the equivalent non-intl-aware function should probably be `strcasecmp`.

zerocrates:
Unfortunate that we have to kind of "double translate" here but I think it's probably unavoidable. Looks good to me.

PR #1759, opened 2021-08-04
Fixed collision with module Annotate.

Daniel-KM:
Fix for the upgrade when the module [Annotate](https://gitlab.com/Daniel-KM/Omeka-S-module-Annotate) is installed.


jimsafley:
This branch just is a proof-of-concept. We're ultimately changing the table name to `value_annotation`, so there's no need to add a migration.

PR #1762, opened 2021-08-22
Enables 'copy' icon.

fukuchi:
This patch enables font-awesome's 'copy' icon. Copy icon is useful for a variety of purposes: this will help module developers.

This patch is a purified version of cc300f468b2661cf7d06dce4faf2575a2733c34d of #1501.


fukuchi:
Here's a typical use-case of the copy icon:
Screenshot from 2021-08-23 11-58-33

zerocrates:
I have a concern with this PR: I'm not sure we want to be providing these "o-icon" aliases for icons we're not actually using.

Modules can of course use any of the FA icons without having this provided in the core's styles just by using the proper `content` styles.

@kimisgold : thoughts?

fukuchi:
@zerocrates,

Thank you for your interest in this pull request. Originally, this PR was part of #1501 and #1503. The simplest way to implement the copy action in Omeka S was to modify Omeka S itself, so I implemented it and submitted the pull requests instead of developing plugins.

That's why I suggested adding copy icon.

PR #1774, opened 2021-11-02
Update _screen.scss

iosdevelop:
In the administration items or media view. The title table is all over the place. The text for the title does not align-left properly. Removing the flex attribute fixes this. This is my first contributing but when I made the fix. I had to copy gulpfile.js and package.json inside the application folder from theme view. After npm install I was able to run gulp CSS and now the items, media, and items set title column are pretty again.

### DIFF
--- a/application/asset/sass/_screen.scss
+++ b/application/asset/sass/_screen.scss
@@ -2945,7 +2945,7 @@ input[type="text"].value-language,
         position: relative;
         width: 100%;
         overflow: hidden;
-        display: flex;
+        display: inline-block;
         align-items: center;
         word-wrap: break-word;
 


iosdevelop:
Screenshot from 2021-11-03 15-41-32

Screenshot from 2021-11-03 15-45-48
)


I made the change _screen.scss


.resource-link {
    display: inline-block;
    align-items: center;
    max-width: 100%;
    min-width: 0;
}

Not .resource but .resource-link was where the fix was. It might not be a thing but I notice this on my instance.


Screenshot from 2021-11-03 15-49-49

I am not sure why it was showing like that.

zerocrates:
So it's primarily that "doubled" one, the bottom one on your screenshot, that you're having a problem with? What does that look like after you've made your change?

It looks to me like the "left" copy of the "Heatstroke, Vol. 7" text is maybe the alt text for a missing image? So maybe that's what really needs to be fixed here?

zerocrates:
Closing this one since we never reproduced it. Feel free to re-file/re-open if there's additional information.

PR #1778, opened 2021-11-10
Fix API link header by setting correct per_page

jimsafley:


zerocrates:
To my reading, the per-page global setting is already correctly applied (it's applied at the time the Paginator is created), so we shouldn't be reading that setting again here.

PR #1781, opened 2021-11-23
Show page private status in navigation when available. (#1744)

kimisgold:


zerocrates:
Some problems here:

- Obviously it needs resolution/rebasing but that's no big deal, it's just the usual compiled CSS conflict.
- Now that the sortable handles are added, this private icon shows up on the wrong side of them
- There's a notice when first loading Navigation when doing this change: the code is looking for an is_public flag on the saved pages and not finding it.
- Related: the is_public data appears to be saved in the tree. Result is that the "eye" does not get updated when the page privateness is changed after it's added to the tree. It's necessary to re-save the navigation to get the correct status. This should instead be working more like the page titles do, which are looked up "live." Fixing this will probably fix the prior bullet.
- The eye doesn't appear either on the buttons to add new pages themselves or in the row added to the tree (it appears after save/reload). The button one is just not implemented but the in-tree one is odd since presumably the actual "is_public" data being set is correct since it's saved in the tree.
- On a stylistic level, the eye is much more prominent in the "link-red" color used in the tree, and for that matter so is the label itself... I suppose this is partially because of the gray backgrounds of the tree rows but do we have a more subdued or neutral option? This is the least important comment obviously.

zerocrates:
The rebase issues got if anything worse after the merge so I force-pushed a rebased version containing (I believe) only the 2 relevant commits. Your local git may complain about pulling this or try to merge it. Deleting the branch locally and checking it out again is probably the best course of action, though `reset --hard` would work also.

I have a deleted "[Missing Page]" page entry in my nav: for one it sort of curiously has the "hidden" eye on it. Clicking edit on that row results in a notice:

> Fatal error: Uncaught Error: Call to a member function isPublic() on null in /var/www/html/omeka-s/application/view/common/navigation-link-form/page.phtml on line 16

I assume that odd eye that's displaying for that missing page is an artifact of the fact that the page was private when I deleted it: leading to a remaining problem from before: the public/private status of a page at the time it was added to the nav still "sticks": if I add a page to the nav when the page was private, and I go and make the page public, if I go back to Navigation the page is still marked private in the tree. Ideally we're not saving that "is this private" data at all and are only displaying it based on if the page is public.

I don't think we need/want the is_public hidden input that's causing the Notice there at all. In the toJstree for the Page type we should be loading the actual page and getting its public status and using that to set the flag, not trying to read that from the saved data.

zerocrates:
Alternate version merged as f89200d524cce32f9de9826b49fbaaa1d233ec3a

PR #1784, opened 2021-11-29
Allow theme config to set server-side required

### LINKED ISSUE #1779: Impossible to use date element in theme config
See https://forum.omeka.org/t/how-to-bypass-required-theme-ini-element-date-field/14292/2

This issue applies to more elements than this of course. We have special handling for some of the well-known and commonly-used problem elements where Zend/Laminas likes to make the element required by default but it looks like we either have to go further there, or provide actual support for theme writers to set required/not-required config elements.

It's possible something like looking for all elements that implement the interface that says they provide a validator and just "fixing" them to not be required could work, or else looking at something like options.attributes.required and setting the input filter accordingly (while defaulting to "not required").

zerocrates:
Makes "required" fields more robust, and fixes usage of elements that
Laminas defaults as required (here we're defaulting to not-required).

(fix #1779)

### DIFF
--- a/application/src/Controller/SiteAdmin/IndexController.php
+++ b/application/src/Controller/SiteAdmin/IndexController.php
@@ -373,18 +373,14 @@ class IndexController extends AbstractActionController
             $form->add($elementSpec);
         }
 
-        // Fix to manage empty values for selects and multicheckboxes.
+        // Set backend required flag according to client-side attr
+        // (also, handle elements that otherwise default to required)
         $inputFilter = $form->getInputFilter();
         foreach ($form->getElements() as $element) {
-            if ($element instanceof \Laminas\Form\Element\MultiCheckbox
-                || ($element instanceof \Laminas\Form\Element\Select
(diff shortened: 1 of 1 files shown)


kimisgold:
This works as expected. I tested using a date element.

1. With no "required" attribute defined, the date element defaulted to not required.
2. Setting `attributes.required = false` remained not required.
3. Setting `attributes.required = true` made the element required, and I could not submit the form until configuring this field.

The one small bump was setting `attributes.required = "false"` with quotes around the value made it set `required="required"` on the element, but that may not be an issue with proper documentation or assumed familiarity with setting boolean values in Laminas.

zerocrates:
Hmm I guess that's just how the ini parser works. Not sure we can really change that. When you wrote it that way it did the whole thing, with the asterisk and all, right?

kimisgold:
> When you wrote it that way it did the whole thing, with the asterisk and all, right?

Yup! Presentation matched functionality.

PR #1787, opened 2021-12-02
Addd block layout template for page title

grandgeorg:
The page title for sites was hardcoded to H2.
Now one can change it to suit their needs.

### DIFF
--- a/application/src/Site/BlockLayout/PageTitle.php
+++ b/application/src/Site/BlockLayout/PageTitle.php
@@ -22,7 +22,7 @@ class PageTitle extends AbstractBlockLayout
     public function render(PhpRenderer $view, SitePageBlockRepresentation $block)
     {
         return $view->partial('common/block-layout/page-title', [
-            'pageTitle' => $view->escapeHtml($block->page()->title()),
+            'pageTitle' => $block->page()->title(),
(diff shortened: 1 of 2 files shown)


zerocrates:
Seems like a perfectly good idea, just to enable theme overriding here.

One note: I think you should pass the title unescaped and then escape it in the view. Just to make anything else a view might want to do with the unescaped title easier, and to align with most of our other usage where the theme does the escaping.

grandgeorg:
Done as recommended.

zerocrates:
Thanks.

Daniel-KM:
There are many things from the module Block Plus (or other modules) that are regularly implemented in core. Is it a good idea if I add some pull requests for some of features? In fact, the perimeter of the core and the modules is not clear. Is there a place where it is defined?

zerocrates:
I don't have a hard and fast rule for you on the boundary of what makes sense in the core vs. modules... I kind of take things on a case by case basis.

I can think of some categories of things that make us think things shouldn't be included, like if they seem like they will be difficult to maintain, or not broadly applicable to users, or incompatible, things like that. But I'm sure there are many improvements to the core's features that are currently in modules that we could/would accept.

PR #1794, opened 2021-12-22
Fix some typos (most of them found by codespell)

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/application/asset/js/global.js
+++ b/application/asset/js/global.js
@@ -290,7 +290,7 @@ var Omeka = {
 
 
     // @see http://stackoverflow.com/questions/7035825/regular-expression-for-a-language-tag-as-defined-by-bcp47
-    // Removes `|[A-Za-z]{4}|[A-Za-z]{5,8}` from the "language" portion becuase,
+    // Removes `|[A-Za-z]{4}|[A-Za-z]{5,8}` from the "language" portion because,
     // while in the spec, it does not represent current usage.
     langIsValid: function(lang) {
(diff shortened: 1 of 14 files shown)


stweil:
Ping?

zerocrates:
Ah yes, sorry. I thought I'd responded to this before.

I think this mostly looks good... who knew we had so many typos in comments. One note: "invokable" is the Zend/Laminas terminology/language when we're using it in comments, so I don't think we'll be changing that to "invocable."

stweil:
I removed the changes for "invocable" now from the pull request.

stweil:
> who knew we had so many typos in comments

That's normal for projects which never used a spell checker like codespell before, so no reason to be worried. :-)

zerocrates:
Thanks.

Oh and sorry, I forgot to mention... I don't know what the situation is with the RDF/XML/OWL files in the application/data/vocabularies folder... those are externally-sourced standardized files and we probably don't want to be fixing their typos unilaterally. Probably best to remove those also.

stweil:
The string _"A foaf:LabelProperty is any RDF property with texual values that serve as "_ comes from `application/data/vocabularies`, but is translated in `application/language`, so I think it has to be reverted there, too.

stweil:
> I think it has to be reverted there, too.

Done now.

zerocrates:
Thanks for making those changes.

PR #1796, opened 2022-01-07
Feature/input search

Daniel-KM:
For accessibility and usage on modern mobile browsers, it is recommended to use input type "search" for search, not "text".


zerocrates:
The conflict is no problem, it's just in the compiled CSS.

I see this changes some of the Chosen override styles... since that markup comes from Chosen it probably doesn't need to be changed, it's going to be a text input regardless.

The addition of `type=submit` to `<button>` elements: by the spec the default type for a `<button>` within a form is already `submit`, so I'm not sure we need to do that part really.

zerocrates:
Closing this on the basis of the "search" type part being done.

If you'd like to do the type=submit one still, feel free to resubmit it, but my comment on that stands, absent some other need for it I haven't considered.

PR #1797, opened 2022-01-07
Used right key for ckeditor4.

Daniel-KM:
In package.json, the right key for "ckeditor" is "ckeditor4".


Daniel-KM:
The check failed, but there is only one line modified in package.json, so you can do it yourself.

PR #1798, opened 2022-01-07
Fix/clean

Daniel-KM:
This is a list of small fixes, code cleaning or small optimizations to be cherry picked.


Daniel-KM:
Tell me if you prefer multiple pr.

zerocrates:
Obviously we're talking about a year-plus ago, but definitely we'd prefer multiple PRs, and that holds true pretty much for any of these multi-topic ones.

Lack of response is on me, but it's much more likely to happen and stay that way when the PR is a bit of a grab-bag like this.

PR #1801, opened 2022-01-07
Added php extensions in composer.

Daniel-KM:


zerocrates:
The allow-plugins key, is that a Composer 2 thing?

Daniel-KM:
Yes, it is installed by default on my computer and it works. I removed them.

zerocrates:
Something about the composer.lock here breaks the test suite... not sure what that's about but some issue from upgrading packages, I imagine. I'll just take the extension changes only at this point.

PR #1802, opened 2022-01-10
Added options to render datatype.

Daniel-KM:
The datatype for value resource append a new argument to render data with the language, but it is not normalized and some other datatypes may need more option than language. 

So I add the argument as an array, to get similar arguments as rendering medias. For compatibility, if the arg is a string in the datatype ValueResource, it's the lang. 

Anyway, this new argument is not used in real life, because  "filter_locale_values" is generally useless, except if locales are exactly the same than the interface, that is rare. Moreover, it requires to update themes to manage a new attribute. After six monthes, no official themes have been updated to add lang except the integrated one. So the compatibility code can be removed in fact.

About language of the value, it is an important thing, and it was also the purpose of the module [Internationalisation](https://gitlab.com/Daniel-KM/Omeka-S-module-Internationalisation) to manage locales similarly, but in the backend, not in the themes: the designers won't have to take care of the language, they have already a lot to do. So the code from #1692 can be modified to remove the option in the views and to check it the in the representation.

### DIFF
--- a/application/src/Api/Representation/ValueRepresentation.php
+++ b/application/src/Api/Representation/ValueRepresentation.php
@@ -44,14 +44,15 @@ class ValueRepresentation extends AbstractRepresentation
     /**
      * Return this value for display on a webpage.
      *
+     * @param array|string|null $options If string, the options is the lang.
      * @return string
      */
-    public function asHtml($lang = null)
+    public function asHtml($options = [])
     {
         $view = $this->getServiceLocator()->get('ViewRenderer');
         $eventManager = $this->getEventManager();
         $args = $eventManager->prepareArgs([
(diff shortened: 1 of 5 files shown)


zerocrates:
I do get the benefit of expanding the lang argument to a more generic options array.

There's a bit of an undesirable effect to adding to the interface's render signature: existing modules that add data types are prone to being incompatible, in a "breaks the site while active" kind of way. This is part of why the initial implementation just tacked `lang` onto the Resource concrete signature without changing the interface.

To avoid this, I think the way to go is to add a new sub-interface that adds the render options argument, with an eye toward a later incompatible release that would just include this in the base interface. As in this branch building on yours: https://github.com/omeka/omeka-s/compare/datatype-options

PR #1804, opened 2022-01-18
Fixed doctrine deprecation warning for using UnderscoreNamingStrategy without making it number aware.

Daniel-KM:


zerocrates:
Huh, the deprecation notice is there in the Doctrine we have but it's silenced with `@` so we don't actually see it.

I wonder if you picked up a newer Doctrine somewhere that actually throws this in a way it can be seen. That would square with the build system problems we saw on your other PR that modified the composer.json. Either that or you have something like XDebug's "scream" on.

Anyway the current locked version also takes this argument and just silently complains about it being missing, so I see no issue taking this change.

PR #1806, opened 2022-01-19
Fix/check extension

Daniel-KM:


zerocrates:
The JP2 part... the media type map array we generate by pulling from Apache's mimetype data, so I don't know if we want to include manual edits there... at that point we'd want to probably modify the scripting that rebuilds the map to either account for our manual additions.

I'll also look a little into why JP2 isn't in the default Apache mime.types, since it's standardized and registered in a way that generally would mean it's included there.

Daniel-KM:
I updated the script to build the list with a specific list.

Note that the first commit is a true real too: if the extension is missing from the url, it should not be checked, since the default extension of the media-type is added later.

zerocrates:
Yes, I caught the extension part of this also. I just want to make sure we're doing things in a reasonable way so that applied default extensions are being checked correctly also, as well as blank-extension files rejected if we can't look up an extension for them.

zerocrates:
So, I checked back up on this. We actually add the "fallback" extension from the media type map beforehand, in `getExtension()`. That's good and makes sense, since it means we're still checking those "guessed" extensions against the whitelist.

But, to my mind, that means your explanation of the change you made here to pass through a missing extension, doesn't do what you described. The check as we have it now should already properly handle extensions that are guessed from the type map.

zerocrates:
I submitted a request up to Apache to include the .jp2 mapping upstream and... there was no response. They're understandably pretty conservative in what they'll change there so your approach of allowing manual additions is probably the way we'd have to go there. JPEG2000 is common enough for Omeka users that it's probably worth accounting for.

zerocrates:
Where I stand currently then, I'd reject the "check of file without extension" aspect of this: files whose extensions were provided from the mimetype map are already allowed with the current code.

zerocrates:
(the commit was a squash of the type map commits and leaves out the "no extension" change)

PR #1808, opened 2022-01-22
Add more events to filter output

jimsafley:
Provides a way for modules to alter the JSON-LD string immediately before API output. Adds the following event:

### rep.resource.json_output

- **jsonLd**: The JSON-LD string.

All classes that implement `Omeka\Api\Representation\RepresentationInterface` trigger this event.

Triggered immediately before API output. To modify the JSON-LD, listeners may modify the `jsonLd` parameter and set it back to the event.

### DIFF
--- a/application/src/Api/Representation/AbstractResourceRepresentation.php
+++ b/application/src/Api/Representation/AbstractResourceRepresentation.php
@@ -286,8 +286,9 @@ abstract class AbstractResourceRepresentation extends AbstractRepresentation
      */
     public function embeddedJsonLd()
     {
-        echo '<script type="application/ld+json">'
-            . json_encode($this)
-            . '</script>';
+        $eventManager = $this->getEventManager();
+        $args = $eventManager->prepareArgs(['jsonLd' => json_encode($this)]);
+        $eventManager->trigger('rep.resource.json_output', $this, $args);
+        return sprintf('<script type="application/ld+json">%s</script>', $args['jsonLd']);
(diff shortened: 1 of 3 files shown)


zerocrates:
OK, I somewhat misunderstood what you were proposing, I was thinking of an event on the Value representations directly.

This is fine, a sort of "last resort" option for modifying the JSON. I think this is also probably doable currently with the Laminas events, though that's not necessarily a reason not to have a more straightforward option like this.

One issue, I imagine this would affect the JSON that's on the API endpoints but not embedded JSON, since it doesn't use the view renderer, etc. Not necessarily an issue in general if documented, but maybe a problem for your intended use case.

jimsafley:
I've added the following events that complement `rep.value.html`:

### rep.value.json

- **json**: The JSON value array

The `Omeka\Api\Representation\ValueRepresentation` class triggers this event. All information about the value is available in the event's target object, including target URL, target ID (for resource value), and label.

Triggered after getting a Value representation's JSON via `jsonSerialize()`. To filter the JSON, listeners may modify the `json` parameter and set it back to the event.

### rep.value.string

- **string**: The value string

The `Omeka\Api\Representation\ValueRepresentation` class triggers this event. All information about the value is available in the event's target object, including target URL, target ID (for resource value), and label.

Triggered after getting a Value representation's string in via `__toString()`. To filter the string, listeners may modify the `string` parameter and set it back to the event.

PR #1811, opened 2022-01-26
Fix some alerts from LGTM

stweil:


stweil:
See https://lgtm.com/projects/g/omeka/omeka-s/ for these and some more alerts.

zerocrates:
We generally use var and not let elsewhere so for consistency those missing declarations should probably be "var"s.

stweil:
I updated the pull request to use `var` instead of `let`.

zerocrates:
After asking for that change I thought again, noting we have a bit of a hodgepodge of approaches on let/const/var already so this probably didn't really hurt any, but thanks for changing it.

PR #1818, opened 2022-02-18
Allow theme to overwrite ckeditor config

grandgeorg:
This is a small but rather important change IMHO, as now one can set theme specific styles etc. for ckeditor (i.e. the default style settings do not make to much sense for a specific theme and maybe somebody wants more or less tools in the toolbar).

### DIFF
--- a/application/src/View/Helper/CkEditor.php
+++ b/application/src/View/Helper/CkEditor.php
@@ -14,7 +14,7 @@ class CkEditor extends AbstractHelper
     public function __invoke()
     {
         $view = $this->getView();
-        $customConfigUrl = $view->escapeJs($view->assetUrl('js/ckeditor_config.js', 'Omeka'));
+        $customConfigUrl = $view->escapeJs($view->assetUrl('js/ckeditor_config.js', 'Omeka', true));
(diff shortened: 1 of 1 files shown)


zerocrates:
I don't think the change you have here will work for that, then. We don't have the theme loaded on the admin where this would be called, so even with "overrides enabled" the theme wouldn't affect the editor toolbar.

Have you tried this out at all?

grandgeorg:
Yes I've tried it, it works fine. I've the config under `themes/my_theme/asset/js/ckeditor_config.js` and it gets loaded in the admin interface instead of `application/asset/js/ckeditor_config.js`.

grandgeorg:
I know that theme stuff interfering with admin the interface could be a concern, but as WYSIWYG is heavily dependent on the frontend I still think it's the right place ...

zerocrates:
We'll have to think more on this... in general the public theme affecting the admin interface like this is unexpected and something we might eliminate... but I take your point on wanting to customize the available styles.

We are adding in the upcoming version a JS event to make it easier for modules to go in and alter the configuration of the editor, which could be an alternative route for doing the kinds of things you're thinking of here.

PR #1819, opened 2022-02-21
Allow to pass return_scalar to the REST API (getList)

jajm:
Sometimes we don't need the full representation of the resource,
sometimes we only want the title or the id.
Having these options available allow fast response times for such cases

A practical use case is this:
- All public items are harvested into another database using module OaiPmhRepository
- New and updated items are kept in sync regularly.
The problem is OaiPmhRepository (or Omeka S) does not keep track of deleted items, or items that go from public to private. Being able to obtain quickly all items identifiers allow to know which previously harvested items have "disappeared" (deleted or made private) from Omeka, and should be deleted from the other database.

With `return_scalar=id` i can query a page of 10000 ids in 1 second. This would take a lot more time without this parameter.

### DIFF
--- a/application/src/Controller/ApiController.php
+++ b/application/src/Controller/ApiController.php
@@ -72,7 +72,15 @@ class ApiController extends AbstractRestfulController
         }
 
         $query = $this->params()->fromQuery();
-        $response = $this->api->search($resource, $query);
+
+        $options = [
+            'responseContent' => $query['response_content'] ?? 'representation',
+            'returnScalar' => $query['return_scalar'] ?? false,
+        ];
+        unset($query['response_content']);
+        unset($query['return_scalar']);
+
+        $response = $this->api->search($resource, $query, $options);
 
         $this->paginator->setCurrentPage($query['page']);
         if (isset($query['per_page'])) {
--- a/application/src/Controller/ApiController.php
+++ b/application/src/Controller/ApiController.php
@@ -74,10 +74,8 @@ class ApiController extends AbstractRestfulController
         $query = $this->params()->fromQuery();
 
         $options = [
-            'responseContent' => $query['response_content'] ?? 'representation',
             'returnScalar' => $query['return_scalar'] ?? false,
         ];
-        unset($query['response_content']);
         unset($query['return_scalar']);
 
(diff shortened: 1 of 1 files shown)


zerocrates:
@jimsafley thoughts?

jimsafley:
I'm in favor of the addition of `return_scalar` param. It's simple and clearly useful. I'm not sure of the usefulness of the `response_content` param. Passing `response_content=reference` will output:

```json
[
    {
        "@id": "http:\/\/localhost\/omeka-s\/api\/items\/97348",
        "o:id": 97348
    },
    {
        "@id": "http:\/\/localhost\/omeka-s\/api\/items\/97352",
        "o:id": 97352
    }
]
```

Which seems not very useful? While passing `response_content=resource` will output:

```json
[
    {},
    {}
]
```

Which is certainly not useful.

jajm:
> I'm in favor of the addition of `return_scalar` param. It's simple and clearly useful. I'm not sure of the usefulness of the `response_content` param. Passing `response_content=reference` will output:
> 
> [...]
> 
> Which seems not very useful? While passing `response_content=resource` will output:
> 
> [...]
> 
> Which is certainly not useful.

I agree. I added it only for "completeness" I guess...

I removed it in the last commit

zerocrates:
My remaining concern here would be if we have situations where there's something in the database that would be accessible with returnScalar that we're omitting from the regular JSON response, and it's something we want to keep private. Things like the password hash for users are obvious examples but there are probably more?

jajm:
Ah, I haven't thought of that... :/

If I understand correctly, the properties "filter" is done by getJsonLd in Representation classes so it cannot be used here.

It might be too much complicated to have a whitelist of properties for each entity, and it would probably break compatibility with modules too (requiring them to implement that list)
Do you have an idea on how to solve this ?

jimsafley:
I have a [proof-of-concept](https://github.com/omeka/omeka-s/compare/return-scalar-query) that will allow `return_scalar` in the query. We're considering it.

zerocrates:
We've merged @jimsafley's alternate approach to this, with the "main" resources enabled at the moment. Further resources could be enabled in the future if requested.

PR #1824, opened 2022-02-28
Add site setting to show/hide value annotations

jimsafley:


zerocrates:
@kimisgold there are probably themes that will need updates to respect this?

kimisgold:
Yup, I updated all the themes with value annotation styles already, but none of them have been released. I'll include this setting before the releases happen.

Issue #1827, opened 2022-03-12
Add arguments to the PageTitle helper to provide control over the CSS classes in the output markup


DanielFlaum:
The [`PageTitle` helper](https://github.com/omeka/omeka-s/blob/e9203959333fdff1fcf6dd9bfa38298ab717a246/application/src/View/Helper/PageTitle.php) currently takes four arguments:

1. The text of the page title
2. The header level to be used in the output
3. Optional text for a subhead label
4. Optional text for an action label

The helper returns a string based on the arguments:

```php
        if ($subheadLabel) {
            $subhead = '<span class="subhead">' . $view->escapeHtml($subheadLabel) . '</span>';
        }
        if ($actionLabel) {
            $action = '<span class="action">' . $view->escapeHtml($actionLabel) . '</span>';
        }

        $level = (int) $level;
        if ($level > 0) {
            return "<h$level>" . $subhead . '<span class="title">' . $view->escapeHtml($title) . '</span>' . $action . "</h$level>";
        }
```
(Although if the header level is specified as `0`, then only the title text is displayed, with no surrounding markup.)

This code includes the hard-coded CSS classes `subhead`, `action`, and `title`. Theme and module authors may need control over these classes. This could be accomplished by adding three additional arguments after the current arguments.

DanielFlaum:
I'm working on producing code for this.

zerocrates:
That's a lot of arguments to add, even though they're optional.

I wonder what you'd think about having this use a partial (which the theme could override) instead?

DanielFlaum:
A partial sounds ideal. I'll go make that now.

PR #1828, opened 2022-03-12
Add arguments for CSS classes to PageTitle helper

DanielFlaum:
This PR adds arguments to the PageTitle helper as proposed at #1827. PHP isn't my forte, so I apologize in advance if I have failed to adhere to any guidelines.

The arguments are added at the end of the method signature, so this should be backwards compatible.


DanielFlaum:
I've switched the PageTitle helper from using new arguments to using a partial.

The reviewer should take particular care in inspecting the test case. I'm not versed in Omeka's test system, so I'm not positive that my test case tests what I think it tests.

zerocrates:
Looks to me like you've lost the headTitle calls which add the elements of the headings to the page `<title>`.

zerocrates:
Also, letting you pass the partial name is fine, but you don't actually need that for the theme to be able to override the markup: it can simply provide a same-named view and its partial will be used rather than the core's.

DanielFlaum:
I passed the partial name by adapting code from another helper already in the source. How about I clean them up in separate PRs?

zerocrates:
I think I'm fine with you keeping the name passing as-is. Just pointing out that you don't need to use that if you want to override this view in a theme.

I do notice the retained call to `$view->escapeHtml` in the partial: in a partial that would generally be `$this->escapeHtml`. Also, you've dropped the escapeHtml calls for the subhead, title and action echo calls: those all need to be escaped as they were originally.

zerocrates:
I think I'm fine with you keeping the name passing as-is. Just pointing out that you don't need to use that if you want to override this view in a theme.

I do notice the retained call to `$view->escapeHtml` in the partial: in a partial that would generally be `$this->escapeHtml`. Also, you've dropped the escapeHtml calls for the subhead, title and action echo calls: those all need to be escaped as they were originally.

Daniel-KM:
Note that like many pull requests, this one is already implemented in a module since a while, in module [Block Plus](https://github.com/Daniel-KM/Omeka-S-module-BlockPlus).

PR #1839, opened 2022-04-17
Fix missing translation in search result

ManOnDaMoon:
Hello,

The site specific search results should display resource titles according to the selected language, if any. This was caused by a missing modification to a `linkPretty` call in the site/index/search view.

Have a nice day

Laurent


zerocrates:
I assume there's the same issue for the no-thumbnail `link` call that's right above there?

ManOnDaMoon:
You're right. I modified the behavior for resources that do not have any public thumbnail to display.

PR #1842, opened 2022-05-20
operator 'contains' and 'equal' for advanced search on user form

ThibaudGLT:
It can be useful to have an operator for advanced search user. A search by email can stay with exact term but maybe it can be helpful to have a "contains" operator for a search by name. I'm a newbie as contributor so I wait your opinion :)


zerocrates:
Hi: can you re-do this pull request so it just contains your changes? It's picking up a lot of our commits also. It looks like this should just be one commit branched off of develop. Rebasing would be one way to do this.

ThibaudGLT:
Oh yes my bad, I rebased it, thanks

zerocrates:
Thank you for the (now long ago) rebase. Sorry for not getting back to you sooner.

The basics here are pretty much fine: these are the right places to add this functionality, you're using the prepared statements right for the SQL query, you have the translation calls you need.

- The name `operator` for your API query parameter would have to change. In context on the page since it's next to the name input it's clear enough what it means, but that context doesn't exist when the adapter looks at the query. It's not obvious from the query key name that this would only affect the `name` query, and we'd have to come up with some pattern for the query names if we wanted to have this same feature on, say, the email field. So something like `name_operator` or `name_query_type` or something would probably be what we'd want here.
- We have an existing pattern for what a "contains" search like this would be called in the API, from the property search for resources. There it's `in` so we'd want to maintain that here, instead of `like`.
- There's some other stuff that would have to change in here: escaping of the query for special LIKE characters, accessibility of your new `select` input (it has no label). Some of that's not perfectly done in the existing code, so it's not a knock on your submission.
- There's some stylistic issues with your code, things like where to break to a new line, how much indentation to use: you're inconsistent between your two `andWhere`s and they should match

I wrote a lot there but nothing there is huge. The bigger questions on our side would be if we want this, if we don't, or if we want it in an expanded form (more query options like are available for properties, offering the option to do the same for email, etc.). We'll get back with more certainty on that point.

PR #1846, opened 2022-06-01
Add a 'no media' option when selecting attachment in page blocks

### LINKED ISSUE #1708: In site pages, anonymous users can see public items without media but not public items with a private media
To reproduce:
1. Create a public item (A) without any media
2. Create another public item (B) with one private media
3. Create a site page with a block 'Item showcase'
4. Add items A and B to the item showcase
5. Logout and go see the page, you should see only item A

I believe that items without any visible media should be treated the same as items without any media.

jajm:
When selecting an attachment to add to a site page block, there is a
"quick add" feature that let the user select multiple items at once.
When this "quick add" feature is used, only the item get attached
(media_id IS NULL is site_block_attachment)

But once the item has been attached, if the user click on "Open
attachment options", and then click on "Apply changes" in the sidebar
without changing anything, the primary media is automatically selected.

So, without changing anything the user actually changed something
without noticing.

Now consider a public item with a single media which is private.
In the first scenario (no media is selected), an anonymous user visiting
the page would see the item title but not the media.
But in the second scenario (a media is selected), an anonymous user
won't be able to see the item at all

And once we are in the second scenario, the only way to go back to the
first is to remove the attachement, and re-add it using the "quick add"
feature.

This patch intends to fix that by providing a "no media" option that is
selected by default (instead of the primary media) when no media was
selected.

Fixes #1708


zerocrates:
There was a failure on the CI but that looks to me like it was a random unrelated issue; I'll have it re-run.

Not sure about the implementation detail here of using a data URI... I think we'd probably do something different there.

I have to do some more looking to see how this is affecting the normal default flow... I think we want to preserve a "choose nothing" option here for the reason you suggest, but may not necessarily want it to be the default.

jajm:
> Not sure about the implementation detail here of using a data URI... I think we'd probably do something different there.

I used a blank GIF I found somewhere because that was the quickest solution I could come up with, but I agree it needs to be changed

> I have to do some more looking to see how this is affecting the normal default flow... I think we want to preserve a "choose nothing" option here for the reason you suggest, but may not necessarily want it to be the default.

Why not ? If this "choose nothing" option is selected, then the primary media will be used, which makes sense as a default behavior in my opinion

jajm:
> I find "No media (primary media will be used, if any)" confusing

That's true. A more accurate description would be "I don't want to choose a particular media, I want Omeka to decide which media to display" but that is a bit long :)
Maybe something like "First visible/available media" would be better ?

> In your branch, it appears that if the user uses the quick add selector and the attachment's primary media is private, the view still outputs an attachment link.

That's what happen currently on Omeka S 3.2.3. When using the quick add selector, the visibility of the attachment depends on the item's visibility, and when not using the quick add selector, the visibility of the attachment depends on the media's visibility.
Speaking of that, I find that is not very intuitive that visibility can depend on the media, in an **item** showcase

PR #1852, opened 2022-06-10
Provide toggle options for blocks in the page editor.

kimisgold:
It can be cumbersome to navigate several blocks on a content-dense page. This branch brings over a feature from ExhibitBuilder in Omeka Classic where users can collapse and expand blocks, both on a per block basis and all at once.


zerocrates:
It's reversible, but I think we're maybe asking for trouble having the collapser/expander sit right next to the "delete this block" button.

zerocrates:
Otherwise, looks good to me.

PR #1853, opened 2022-06-12
avoid indexing browse and search pages

raboof:
Search pages and browse/search results pages are 'light on content', and might be considered 'low-quality' by search engines.

By explicitly tagging these pages as 'not to be indexed', they are not counted when determining the 'overall site quality', which should improve the overall site quality (and thus ranking) of Omeka S-based sites.


zerocrates:
Though it doesn't disable crawling, I'd be concerned about `noindex` on browse pages in particular causing less crawling, and therefore less discovery of the actual resource show pages that you _do_ want indexed.

PR #1860, opened 2022-06-24
IIIF presentation media & /iiif-viewer endpoint

jimsafley:


zerocrates:
Dimensions is always an interesting question. We have this same kind of situation for other similar viewers like the Openseadragon one. Media render can pass through options so probably the best thing would be to respond to height (and possibly width) options. Themes could then use that currently and if we later plumbed in some easier ui for users to change that it would be already there and ready to work.

As for options to the viewer... my initial concern would be security... are there things we wouldn't want to be so easy to set with a query string like that? I'm not sure.

PR #1866, opened 2022-07-08
Add option to make asset URLs absolute

jimsafley:
Adding the option to make asset URLs absolute will make methods like `MediaRepresentation::thumbnailUrl()` have more consistent return values.


zerocrates:
With it being used so many places, I think it's prudent to just change the `$basePath` variable once up top if the absolute flag is set, rather than check it in all the places its being used. Then you also wouldn't have to even call serverUrl() at all when an absolute URL isn't being requested.

zerocrates:
Rebased and merged.

PR #1874, opened 2022-07-29
Improve helptext for Site thumbnail

triplingual:
As far as I can see, the thumbnails will always be square, so the helptext should make a note of that.

### DIFF
--- a/application/src/Form/SiteForm.php
+++ b/application/src/Form/SiteForm.php
@@ -51,7 +51,7 @@ class SiteForm extends Form
             'type' => 'Omeka\Form\Element\Asset',
             'options' => [
                 'label' => 'Thumbnail', // @translate
-                'info' => 'Choose or upload a thumbnail to display with site.', // @translate
+                'info' => 'Choose or upload an image to display with site. The image will be cropped to a square if not already square.', // @translate
             ],
             'attributes' => [
                 'id' => 'thumbnail',


zerocrates:
We don't actually crop anything here (Assets are never edited when uploaded), but there can be CSS that's having a similar effect for display purposes.

My concern is that this squaring-off might not be consistent everywhere the thumbnail might be displayed (especially in the "list of sites" block since its appearance can be controlled by the theme).

sharonmleon:
Definitely have rectangular site assets in my list of sites block in my [test install](http://6floors.org/workspace/omeka-s/s/base/page/welcome)

triplingual:
In part I'm relying on the use of "square" in [the site-list-enry.html file](https://github.com/omeka/omeka-s/blob/91cda586589a69dc37b9b9ca30a4136a9c2afe97/application/view/common/site-list-entry.phtml#L7) for my assertion about thumbs always being "square". Is "square" a term of art in that case that includes "rectangle"? (And as I recall from my school geometry, a square is a rectangle but a rectangle is not a square. 😄 )

zerocrates:
I can't dispute your mastery of geometry. The trick here is that "Asset" thumbnails always ignore that parameter. "square" and "large" and so on only have meaning for generated thumbnails from media. We use the same thumbnail helper even if there's no possibility of having media-derived thumbnails (like here), but it still requires the size parameter to be passed.

This is a good nudge to me to look at cleaning that up so we can pass null here (both where it's irrelevant and where we'd like to make the actual choice of thumbnail configurable).

triplingual:
Ah! All makes sense now. Upvote this comment and I'll close, or you can close.

PR #1879, opened 2022-08-09
Add language to URI data type

### LINKED ISSUE #1060: Language codes for URI reference labels?
Per https://forum.omeka.org/t/internationalized-omeka-s/3154/19, it might be nice to allow i18n on the labels for URI references in metadata, similar to what we have for literals.

jimsafley:
Includes new styling for language control. Fixes #1060


zerocrates:
@jimsafley I've merged this with some fixups to the styling around language "active"ness (and some further brokenness on the value annotation input)

Note: this needs to be extended to the value annotation input for URIs.

Issue #695, opened 2016-07-19
On active browse page, all other browse pages active in navigation


kimisgold:
If there are multiple browse links in a site's navigation, landing on one will mark all browse links in the navigation as active in the public views.

zerocrates:
Are they all exactly the same? If so there's not much we can do about it: active-ness is judged by whether the routing parameters match and multiples of the same link will always all match.

kimisgold:
They are all 'browse' custom links using different labels and queries.

jimsafley:
The only way I see this working is changing the Browse navigation link from a MVC page to a URI page, like this:

```php
// Omeka\Site\Navigation\Link\Browse::toZend()
public function toZend(array $data, SiteRepresentation $site)
{
    $urlHelper = $this->getViewHelperManager->get('url');
    return [
        'type' => UriWithQuery::class,
        'uri' => $urlHelper(
            'site/resource',
            ['site-slug' => $site->slug(), 'controller' => 'item', 'action' => 'browse'],
            ['query' => $data['query']],
        ),
    ];
}
```

This will require injecting the view helper manager in to the Browse navigation link via a factory. It will also require extending the `Uri` page class with `UriWithQuery` to override the `isActive()` method, like this:

```php
// Omeka\Navigation\Page\UriWithQuery
public function isActive($recursive = false)
{
    $uri = $this->getRequest()->getUri();
    if (! $this->active) {
        if ($this->getRequest() instanceof Request) {
            if ($uri->getPath() . '?' . $uri->getQuery() === $this->getUri()) {
                $this->active = true;
                return true;
            }
        }
    }

    return parent::isActive($recursive);
}
```

Note that this now appends the query string to the page's URI path, so it matches with the current URI, which makes the current navigation link active, but none others. Admittedly, this solution is a bit hacky, but, as far as I know, it's the one that has the least impact.

PR #1880, opened 2022-08-09
Mark browse navigation links as active only when paths and queries match

jimsafley:
Before, only paths needed to match, making every browse link in navigation active, regardless if they had distinct queries. (#695)

### DIFF
--- a/application/src/Site/Navigation/Page/UriWithQuery.php
+++ b/application/src/Site/Navigation/Page/UriWithQuery.php
@@ -2,6 +2,7 @@
 namespace Omeka\Site\Navigation\Page;
 
 use Laminas\Http\Request;
+use Laminas\Navigation\Page\AbstractPage;
 use Laminas\Navigation\Page\Uri;
 use Laminas\Uri\UriFactory;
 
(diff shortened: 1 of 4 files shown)


application/src/Site/Navigation/Page/UriWithQuery.php
```diff
@@ -0,0 +1,36 @@
...
+                $uriCurrentQuery = $uriCurrent->getQueryAsArray();
+                unset($uriCurrentQuery['page'], $uriCurrentQuery['sort_by'], $uriCurrentQuery['sort_order']);
+                $uriCurrent->setQuery($uriCurrentQuery);
+
+                $uriPage = UriFactory::factory($this->getUri());
+                $uriPageQuery = $uriPage->getQueryAsArray();
+                unset($uriPageQuery['page'], $uriPageQuery['sort_by'], $uriPageQuery['sort_order']);
+                $uriPage->setQuery($uriPageQuery);
+
+                $identicalPaths = $uriCurrent->getPath() === $uriPage->getPath();
```
zerocrates, line 25: Slight concern about this overmatching... it wouldn't come up in the context its being used here for just browse pages, but the original Uri compares the request path to the full uri set to the nav page, which accounts for making things like absolute uris never "active" as they'd never match. Here if an absolute URI uses this class it would be "active" as long as the paths match even if it has a scheme/host/etc.

Issue #1695, opened 2021-03-12
Select All in Attachment sidebar 


georgiawebgurl:
Consider as an enhancement to the Attachment sidebar:
Change ‘Quick add’ to ‘Select multiple’ and add a button to ‘Select all’ 

This came out of a discussion about adding multiple images in the carousel browse and what might be an more efficient workflow. 
Thanks!

kimisgold:
This feature is ready for a look on the `select-all-sidebar` branch.

PR #1883, opened 2022-08-11
Include "select all" button for resource selector.

kimisgold:
This addresses #1695. The resource select sidebar now has a "select all" button. It behaves as follows:

* In its initial state, it is inactive and appears with an empty checkbox. Screen readers will read it as "Select all is inactive". 
  * When the user clicks on its initial state (unchecked), the "quick add" checkboxes open up, and all the checkboxes are ticked. Screen readers will read a success message that says "Quick add checkboxes are now available. All results are now selected."
  * The button itself then displays a ticked checkbox icon. When focused, screen readers will read it as "Select all is active."
* If the user clicks the active "select all" button and no individual resource has been unchecked, it will uncheck all boxes. The "select all" button returns to its initial inactive state.
* If the user deselects a resource while the "select all" button is active, it goes back to its inactive state.
  * Clicking the "select all" button again immediately after will again select all resources and return the button to its active state.
* If the user toggles off the "quick add" interface while all resources are selected, the "select all" button will remain in its active state, indicating that even if the checkboxes are not visible, they are checked.


application/view/omeka/admin/item/sidebar-select.phtml
```diff
@@ -64,13 +64,18 @@ $expanded = $resourceClassId || $itemSetId || $id;
     </div>
 
     <?php if ($itemsFound): ?>
-    <button type="button" class="quick-select-toggle"><?php echo $translate('Quick add'); ?></button>
+    <span class="sr-only quick-select-on success" tabindex="-1"><?php echo $translate('Quick add checkboxes are now available.'); ?></span>
+    <span class="sr-only quick-select-off success" tabindex="-1"><?php echo $translate('Quick add checkboxes are now hidden.'); ?></span>
+    <button type="button" class="quick-select-toggle"><?php echo $translate('Quick add'); ?> <span class="sr-only active"><?php echo $translate('Is active'); ?></span><span class="sr-only inactive"><?php echo $translate('Inactive'); ?></span></button>
+    <span class="sr-only select-all-on success" tabindex="-1"><?php echo $translate(sprintf('%s All results are now selected.', $translate('Quick add checkboxes are now available.'))); ?></span>
```
zerocrates, line 70: Reversed order of the translate and sprintf calls here (the first or "outer" one specifically).

application/view/omeka/admin/item/sidebar-select.phtml
```diff
@@ -64,13 +64,18 @@ $expanded = $resourceClassId || $itemSetId || $id;
     </div>
 
     <?php if ($itemsFound): ?>
-    <button type="button" class="quick-select-toggle"><?php echo $translate('Quick add'); ?></button>
+    <span class="sr-only quick-select-on success" tabindex="-1"><?php echo $translate('Quick add checkboxes are now available.'); ?></span>
```
zerocrates, line 67: It seems to me that these two spans are always screenreader-visible, am I right on that? Seems like they could get read outside of being explicitly focused by the JS.

application/asset/js/resource-selector.js
```diff
@@ -52,8 +53,34 @@
         });
 
         $('#select-resource').on('click', '.quick-select-toggle', function() {
-            $(this).toggleClass('active');
+            $(this).toggleClass('active').find('button .sr-only').toggleClass('active');
```
zerocrates, line 56: Little odd to toggle the "active" class on the screenreader text inside the button but not the "inactive" class... I don't quite get why this is necessary. You have a comment on the corresponding styles but I'm not getting the distinction.

application/view/omeka/admin/item/sidebar-select.phtml
kimisgold: That's a good point, as I was only testing via tabbing through. They should probably have `display: none` when not active.

application/asset/js/resource-selector.js
kimisgold: I don't think it's necessary after all. I think these classes and their toggling got away from me, gonna pare them down.

zerocrates:
My testing indicates this breaks the loading spinner for the sidebar: since this adds the "regular" weight fa icons, and the style for the loading spinner asks for regular weight. It really needs bold, but previously when we only had the bold weight ones, it worked. Fix is simply just asking for bold in that style, but worth a little search through to see if the same thing is being done elsewhere.

Separate problem: since this was created we merged in a feature to extend quickadd to the item set sidebar also, so this should be used there also.

PR #1899, opened 2022-08-25
Set custom element values to strings on error

jimsafley:
These two elements are custom built in common/property-form-batch-edit.phtml
and will POST arrays. The form knows about the elements via hidden inputs,
which expect strings. If the form has an error, we must set the input values
to empty strings or there will be errors.


zerocrates:
This probably calls for an in-code comment... we'll never remember what those lines are for later.

PR #1902, opened 2022-08-29
Add page date/time block

jimsafley:


application/view/common/block-layout/page-date-time.phtml
```diff
@@ -0,0 +1,16 @@
+<?php
+$page = $block->page();
+$created = $page->created();
+$modified = $page->modified();
+$dateFormat = $block->dataValue('date_format');
+$timeFormat = $block->dataValue('time_format');
+$siteLocale = $this->siteSetting('locale');
```
zerocrates, line 7: We set the default locale configured when loading the site, so you shouldn't need to look up and pass one explicitly.

It also might cause issues to do so (if there is no locale selected, for example?)

PR #1912, opened 2022-08-31
Show only item sets in site on item show page

jimsafley:
See #1724


zerocrates:
This has to be resolved to cover Kim's later changes to the elements used, but that's no big deal.

I guess the only question on this one is if this rises to the level of wanting/needing an option to enable/disable this behavior. I'm open to it just being always-on (or driven by the same setting as your other related PR about values linking outside the site pool?).

PR #1958, opened 2022-10-12
Added the resource to output of displayValue().

Daniel-KM:
Often in template display-values, the resource is needed to manage theme specificities, but it is not available directly.


zerocrates:
Looks good to me, avoids having to pick out a value and grab the resource from that or whatever you'd have to do otherwise.

PR #1960, opened 2022-10-13
Exclude resources not in site & improved linked resources

jimsafley:


application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -451,17 +451,42 @@ public function objectValues()
...
     public function displayValues(array $options = [])
     {
         $options['viewName'] = $options['viewName'] ?? 'common/resource-values';
-        $partial = $this->getViewHelper('partial');
+        $options['siteId'] = $options['siteId'] ?? null;
+
+        $services = $this->getServiceLocator();
+        $values = $this->values();
+
+        if ($options['siteId'] && $services->get('Omeka\Settings\Site')->get('exclude_resources_not_in_site')) {
```
zerocrates, line 467: This bit is a little hinky: it checks for a passed in site id but then checks against the current site's settings. Better probably to move the check out into the partial so you only pass the site id if the current site is configured to exclude.

This change would allow for what the siteId option would seem to imply, that you can pass a different site's id and have it work right, or that you could pass one from the admin side.

PR #1967, opened 2022-10-31
Improved performance of display of the item page.

Daniel-KM:
When the owner is loaded, it can imply many things useless in form.

### DIFF
--- a/application/view/omeka/admin/item/manage-item-sets.phtml
+++ b/application/view/omeka/admin/item/manage-item-sets.phtml
@@ -13,11 +13,8 @@ if ($item) {
 
 $itemSets = [];
 foreach ($itemSetRepresentations as $itemSet) {
-    $owner = $itemSet->owner();
     $itemSets[] = [
         'id' => $itemSet->id(),
-        'title' => $itemSet->title(),
-        'email' => $owner ? $owner->email() : null,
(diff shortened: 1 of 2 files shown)


zerocrates:
This seems like a fine change to me. Actually we could go even further: the only part of this array that's actually used is the `id` key: we're now just looking up the data to use in the main table by pulling from the selector sidebar. So we don't need the title and email keys at all.

I'm curious: do you get a noticeable performance improvement out of this? I'd be kind of surprised, but I suppose with lots of saved sites you're potentially making a very big JSON. Either way we don't need all the extra junk so it should be removed, but I'm interested in what effect you've seen from this change.

Daniel-KM:
Yes, it can be pushed on branch 3.2, not on dev, but i don't know where is the roadmap.

Yes, the performance is significant in one of my sites, but related to another module, where I did a `json_decode(json_encode($user), true)`, that allowed to access quickly to the full array of a user (not `$user->jsonSerialize()`, that does not encode sub-resources), but there were slow events on users and related data. There is probably something to do to get only references, not representations. 

Normally, I removed the json_decode/encode resource construction of my modules where it had bad impact.

Daniel-KM:
Updated to keep only id.

PR #1969, opened 2022-11-11
Increased minimal database versions to mysql/8.0.18 and mariadb/10.2.27.

Daniel-KM:
Hi,

For next major version, it's probably time to increase the minimal version of the database, like other parts of the environment (php 7.4, 28 nov 2019). Current minimal versions are mysql 5.6.4 (20 dec 2011) and mariadb 10.0.5 (7 nov 2013). 

The versions available in nov 2019 are mysql 8.0.18 and mariadb 10.2.27. So this is the minimum right version to choose. 

In common distributions using php 7.4 by default, there are debian 11 (14 aug 2021) with mariadb 10.5.11 (no mysql ), ubuntu 20.04 lts with mysql 8.0.19 / mariadb 10.3.22, centos 8 / mysql 8.0.26 / mariadb 10.3.28. So, they are the optimal versions to choose.

There are a lot of new feature, in particular optimized json storage and json queries and spatial sql. And the mysql ad says that "mysql 8 is [up two times faster](https://www.mysql.com/products/enterprise/database/)". 

Note that mysql 8.0.18 and mariadb [10.2](https://mariadb.com/kb/en/changes-improvements-in-mariadb-102/) are no more supported. But this is the same for php 7.4 too ([until 28 nov 2022](https://www.php.net/supported-versions.php)).

So this proposition is very conservative, but something more recent will be better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ See the [user manual](https://omeka.org/s/docs/user-manual) for more information
 ### Requirements
 * Linux
 * [Apache](https://www.apache.org/) (with [AllowOverride](https://httpd.apache.org/docs/2.4/mod/core.html#allowoverride) set to "All" and [mod_rewrite](http://httpd.apache.org/docs/current/mod/mod_rewrite.html) enabled)
-* [MySQL](https://www.mysql.com/) 5.6.4+ (or [MariaDB](https://mariadb.org/) 10.0.5+)
+* [MySQL](https://www.mysql.com/) 8.0.18+ (or [MariaDB](https://mariadb.org/) 10.2.27+)
 * [PHP](https://www.php.net/) 7.4+ (latest stable version preferred, with [PDO](http://php.net/manual/en/intro.pdo.php), [pdo_mysql](http://php.net/manual/en/ref.pdo-mysql.php), and [xml](http://php.net/manual/en/intro.xml.php) extensions installed)
 
 ### Generating thumbnails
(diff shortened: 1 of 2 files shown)


zerocrates:
In general, I'd rather not increase the minimums for things like this unless we're actually taking advantage of something that the current minimums don't allow for. I'm not aware of anything we're really itching to use that needs newer SQL server versions, but I'm open to hearing about possibilities there of what we (or module writers) might want to use.

We purposely don't collect analytics or telemetry from installations, so I don't have a handle on what usage out in the wild is like but my instinct says that we almost certainly have a _lot_ of users on MySQL versions prior to 8, as 8 was considered a major change that people have probably been a little reluctant to jump on. Moving to 8 also takes us into some more dangerous waters, as the differences between MySQL and MariaDB have grown in their more recent versions. In particular, MariaDB 10.2 is more of a match feature-wise for MySQL 5.7.

We tend to be a little more aggressive about moving the PHP minimum, partially out of necessity because our dependencies do as well, and noting an increased trend in the world for even shared hosts to be on more bleeding-edge PHP versions than was previously the case. Neither really apples so much to databases, though.

zerocrates:
Since we did our variant of this one that's lighter on the MySQL side, we won't take this particular PR.

We're still open to hearing arguments for further increases for the database minimums in future versions, though as I mentioned here, the decisionmaking does get a little trickier once we get into MySQL 8 territory.

PR #1981, opened 2023-01-06
Fixed site-page-edit.js.

Daniel-KM:


zerocrates:
What's the bug?

Daniel-KM:
For the block Asset, the button may not work (the selector hidden shuold have a "]" and there may be annother .selector-asset on the page.

zerocrates:
Do you happen to know of a situation where there is another .selected-asset? Just so we can confirm the fix. Another block, presumably, perhaps from a module?

I'm wondering if we might want to tie this not just to sidebars generically but to the asset sidebar specifically, `#asset-options`.

Daniel-KM:
When i updated module [Block Plus](https://github.com/Daniel-KM/Omeka-S-module-BlockPlus/commit/9d0b823ccb5bd19c2c38e7bc312f74a369e5a3f7) in March, it didn't work and it was the fix. Maybe it is useless now. I'm finishing upgrading the module, so i let you know on Monday if it is still needed.

Daniel-KM:
I updated the module and i don't need the class .sidebar-content any more for this. In previous version, the js was overridden.

zerocrates:
So, are you saying you no longer need that part of the PR? I'm happy to have it, though I'd probably prefer it as `#asset-options .selected-asset` if we're going to make a change there.

PR #1987, opened 2023-01-18
Add batch edit owner

### LINKED ISSUE #1985: Batch edit owner
There's no batch edit option to change the owner of a resource. This could be useful for admins dealing with deleted users or coming from contexts where no owner is set by default like Collecting.

jimsafley:
Fixes #1985


zerocrates:
I'm thinking we want to check that the user has the necessary privilege before deciding to show this input?

The actual job will check it but we shouldn't show the option if we know it won't work.

zerocrates:
The actual underlying ACL check for whether a user can do this is on `change-owner` against the entity, probably best if we're checking that specifically.

PR #1988, opened 2023-01-21
Add direct link to current user settings

### LINKED ISSUE #1976: More direct way to get to your own user settings
You currently have to go through your user show by clicking your name in the main nav, then edit your user, then go to the settings tab. Alternatively you can go to user browse, pick to edit yourself, and continue in the same way. Both are a little non-obvious and require a few clicks and also even noticing that the Settings tab is there.

We probably want some more direct route to get to the settings for a user. Maybe just another link in the main admin nav, or even a shortcut from the global settings to your user's settings. If only to clearly indicate that there _are_ user settings.

jimsafley:
Fixes #1976


zerocrates:
@kimisgold @allanaaa thoughts here?

This adds a link to the current user's settings (since as mentioned in the linked issue #1976 we have more user setting-y stuff so you might want to go there more often and it's currently a several-click route). The link would appear in the "Admin" section of the left nav, as "My settings."

My initial thought is: we have a fair amount of space up top near the Logout button and current user's name, and would we maybe prefer it up there? As a gear icon (and/or edit pencil icon if it would go to just Edit)?

allanaaa:
Yes, I think near the username and the logout button would be a good spot.

kimisgold:
Sounds good to me. I think I lean a little towards gear for configuration, but I can style that once it's merged.

jimsafley:
@kimisgold It's ready for your styles.

zerocrates:
I think the single-line "login status" the new styles here use is fine... though the "logged in as" text gets a little awkward this way. Would we maybe prefer just the "user" icon in white there instead of the text? Gives us some more space for longer user names to fit in the single line, also

PR #1997, opened 2023-02-03
Fix/value annotation resource

Daniel-KM:
The resource value annotations use the same template than resource values, so it should pass the resource too.

By the way, there is an important missing feature about value annotation : there is no template, so the user should select the properties and the datatype one by one. Most of the time, it's only one or two properties, but it should be listed somewhere, so it can be filled in one click.

So is it better I improve the module Advanced Resource Template for that or to push request here to avoid a future re-implementation, because it is probably something people will want later. 

As for many other things, is there a way to know what will be inside the core and what will remain in the modules to avoid to develop the same thing multiple times with subtle variants?


Daniel-KM:
Sorry, the commit was passed against master instead of develop and I can't change it here.

zerocrates:
Probably best to just make a new PR against develop with the single commit at issue here, and we'll close this one.

PR #1999, opened 2023-02-06
Passed resource and value annotation to template for displayValues().

Daniel-KM:
The resource value annotations use the same template than resource values, so it should pass the resource too.

By the way, there is an important missing feature about value annotation : there is no template, so the user should select the properties and the datatype one by one. Most of the time, it's only one or two properties, but it should be listed somewhere, so it can be filled in one click.

So is it better I improve the module Advanced Resource Template for that or to push request here to avoid a future re-implementation, because it is probably something people will want later.

As for many other things, is there a way to know what will be inside the core and what will remain in the modules to avoid to develop the same thing multiple times with subtle variants?

### DIFF
--- a/application/src/Api/Adapter/ResourceAdapter.php
+++ b/application/src/Api/Adapter/ResourceAdapter.php
@@ -33,7 +33,6 @@ class ResourceAdapter extends AbstractEntityAdapter
      *
      * This version simply proxies to the "real" getRepresentation for each resource's adapter.
      *
-     * @param string|int $id The unique identifier of the resource
      * @param mixed $data Whatever data is needed to compose the representation.
      * @return ResourceInterface|null
      */
--- a/application/src/Api/Adapter/ValueAnnotationAdapter.php
+++ b/application/src/Api/Adapter/ValueAnnotationAdapter.php
@@ -2,7 +2,9 @@
 namespace Omeka\Api\Adapter;
 
 use Omeka\Api\Representation\ValueAnnotationRepresentation;
+use Omeka\Api\Representation\ValueRepresentation;
 use Omeka\Api\Request;
+use Omeka\Api\ResourceInterface;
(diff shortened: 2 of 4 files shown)


zerocrates:
Probably the question about templates applying to the annotations should be an issue of its own, it'll get lost here, even moreso than other things.

On the question of what's in the core and what's in modules, that's probably also a separate issue and if we want to discuss it at any length we should make one for that as well. My short answer is that we of course don't want unnecessary duplication of effort between the core and modules, not only is this unfortunate on its own terms for the developers of both, it's confusing for users. 

Some duplication does end up making sense though: modules can demonstrate user interest and needs, and the feature can ultimately make sense being there by default, or it can be easier if it's there for other modules to rely on, or various other reasons. AltText is an example of something I myself did as a module before adding a very similar feature to the core later, similarly we've long had an unreleased module for manually setting the server base URL that I have an eye toward (eventually) moving to the core.

In terms of planning: there's a recently added, very high-level, list of "priorities" in the [S dev docs](https://omeka.org/s/docs/developer/devpriorities/). Of course, our own issues and branches for actual in progress work are openly available here on Github. And coming from the reverse direction, we're happy to get issues and questions coming in about the interest or likelihood of us supporting particular features. In particular, features and setups that allow modules to more easily and reliably accomplish what they need to are something I'm interested in even if we don't want the module itself in the core.

Some of this kind of dialogue we've had in the past and it hasn't always been effective for one reason or another. We pretty recently did a fairly thorough clean-out of the old issues on the tracker with an eye to making sure we wouldn't lose track of things, so please feel free to re-raise an issue if it had gotten lost before.

Daniel-KM:
Ok, I implemented the feature "template for value annotation" in module [Advanced Resource Template](https://github.com/Daniel-KM/Omeka-S-module-AdvancedResourceTemplate) v3.4.22.
For the current pr, it's still useful.

PR #2001, opened 2023-02-07
Added system info about free space.

Daniel-KM:
Space issue is not so rare, so to give information about it may be interesting.
Partially integrate https://github.com/symac/Omeka-S-module-FreeSpace


zerocrates:
I think we're happy to take this, though I won't move forward on any new feature PRs just yet until we have 4.0.1 out the door.

zerocrates:
Slight issue: we use the "binary" definition of megabyte when reporting the max filesize for upload in the UploadLimit helper. We might want to keep that consistent here.

Daniel-KM:
Ok, i pushed the fix.

PR #2002, opened 2023-02-07
Fix batch edit selected items

loyolakain:
There is issue when we bulk edit selected items with some modules: the process breaks.


zerocrates:
I think we'd need to see some concrete cases of this problem happening before accepting any fixes for it.

In particular, in-process batch edits (so, "batch edit selected" ones) already don't do the "detach" process that's generally the cause of these problems. So I'm hesitant to introduce a `clear()` where we shouldn't need to have one.

There's a reference in the commit messages to this problem occurring with NumericDataTypes: do you have a reliable way to reproduce the problem?

loyolakain:
Yes, I still have the problem, with omeka s version 4.0.0, with NumericDatatypes as the only module, if in the batch action I chose "Edit selected" instead of "Edit all", the "Clear property values" batch does nothing, the values aren't removed for the properties I have chosen.

zerocrates:
Thanks, this minimal description of how to reproduce the problem is really helpful. We're pretty sure we've zeroed in on the issue and have a fix in progress that fixes it in a slightly different way than this PR does.

One question just to confirm: is it _specifically_ the "clear property values" batch edit that you've been having problems with? The fix we have specifically is a problem with that action (and the visibility action).

loyolakain:
Yes, "the clear property values" is the action not working

zerocrates:
OK, great. That matches up with the fix we have in the works, which specifically relates to that batch action.

I'll update this thread when that fix goes in; ultimately we'll probably close this PR but the bug should be fixed.

zerocrates:
The fix I was talking about was just merged, it was #2031. Instead of using `clear` between runs of the batch edit, it alters the internals used for "clear property values" and one other action.

The merge also included some other aspects and small improvements addressed by your PR in a second commit.

Daniel-KM:
There are two more issues that were not included in #2031 : 

* [Cleaned list of actions to process.](https://github.com/omeka/omeka-s/pull/2002/commits/7139a16d730fc0574ca0c013aa555a221b327399)
* [Made the behavior of batch edit selected similar to batch edit all.](https://github.com/omeka/omeka-s/pull/2002/commits/6422dd036b792c98f13c02228f3c5c0b5859af97)

Issue #1998, opened 2023-02-03
Site option to expand value annotations by default


jimsafley:
See this [forum post](https://forum.omeka.org/t/value-annotation-page-design/16625/1). It's reasonable to expect a site setting that allows users to toggle whether they want value annotations to be open by default, without a click.

PR #2010, opened 2023-02-14
Add site setting to expand value annotations on page load

jimsafley:
See #1998


zerocrates:
It's clever to have this work with the old values, but I'm thinking we'd probably rather migrate them and then take out the special handling in the code (even though it's small).

PR #2013, opened 2023-02-20
Fixed display of tabs on resource edit page when tabs are numerous.

Daniel-KM:


kimisgold:
From what I can tell, the tabs are already wrapping in this view. Adding flexbox creates and odd overlap with the main form. 

image

Entirely possible I'm missing something. Could you share screenshots of the problem you're trying to solve alongside how your solution addresses it?

Daniel-KM:
It works for me. In fact, the issue is not on edit page, but on show page.
Capture d’écran du 2023-02-27 10-53-08

kimisgold:
I think setting `clear: both` on `.active.section` might get the same result without causing more issues on the edit view.

Daniel-KM:
OK.

PR #2014, opened 2023-02-20
Escaped value when searching with "like" for contains, start with and end with.

Daniel-KM:
… end with.


zerocrates:
Looks good to me; thanks.

PR #2017, opened 2023-02-22
Enable search operation for API resource adapter

jimsafley:
See this forum post https://forum.omeka.org/t/query-multiple-resources-by-ids/16833

I don't remember if there was a reason why we disabled the `search` operation for the ResourceAdapter. Also, I don't remember if there was a reason why we enabled the `batchUpdate` operation, or if it was an oversight.


Daniel-KM:
This is included in PR #1799, that managed create too.

zerocrates:
@Daniel-KM The two PRs do the "search" aspect very similarly, but Jim's has a couple advantages over yours in a couple small places, so I'm taking this one.

As for the other operations you included in your PR, I'm not as convinced about adding those, I'll close #1799 just for clarity of what's going on, but if you'd like you can include just your changes to add the other operations and we can discuss in a new PR for that.

PR #2022, opened 2023-02-27
Removed link to query-form.css (fix #2015).

### LINKED ISSUE #2015: Query Form css is empty
I noticed this file https://github.com/omeka/omeka-s/blob/develop/application/asset/css/query-form.css is referenced in the code but is empty, which gives a 404 error on editing a page. Happens in v.3.2.x and v4 it looks like.

Thanks,

Joseph

Daniel-KM:


zerocrates:
Thanks; we actually caught also another missed instance of this so we made a commit cleaning up both of them.

PR #2023, opened 2023-02-27
Feature/scalar fields more

Daniel-KM:
Some entities have not available fields to get return via scalar fields. It's interesting in particular for site pages, but it can be improved for all of them. Or is there a reason to limit this feature to item, item set and media?


zerocrates:
I don't think there's any particular reason behind this just having been done for the "main" resource types, other than that those are what we're often most interested in, at least to start.

PR #2042, opened 2023-03-28
Fixed routing when site-slug is boolean false.

Daniel-KM:
Because in mysql, 
```sql
SELECT * FROM site WHERE slug = 0 LIMIT 1;
``` 
returns something.

### DIFF
--- a/application/src/Mvc/MvcListeners.php
+++ b/application/src/Mvc/MvcListeners.php
@@ -293,9 +293,7 @@ class MvcListeners extends AbstractListenerAggregate
 
         $event->getViewModel()->setTemplate('layout/layout-admin');
 
-        if ($routeMatch->getParam('__SITEADMIN__')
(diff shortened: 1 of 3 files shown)


zerocrates:
Yeah, that's definitely a problem.

Your solution as written has an issue if the slug is string that's boolean false though, in particular if it's the string `0`. (We have a separate problem of a similar plain boolean check on the slug in `prepareAdmin` in this same class that I noticed when checking on this.)

I think we're better off here with a check like `!(is_string($siteSlug) && strlen($siteSlug))` for the quick return (`=== null` would work but I'd like to have it pick up empty-string also so I think the strlen option is clearer). With that in place (and the existing boolean check of the slug in prepareAdmin removed) we'd fix this issue and also correctly allow for the slug `0`.

zerocrates:
Thanks.

PR #2044, opened 2023-03-29
Added a message for performing job.

Daniel-KM:
Avoid mysterious issue.


zerocrates:
So, this is addressing a situation where users try to run a job but the MySQL PDO driver isn't installed on the CLI version, I take it.

What error do they get absent this change?

I'm a little concerned about trying to match the string of the error message, and whether that could change, or use a different language based on system locale, etc. If this is common and confusing enough that we need a specific message for it, we might just want to explicitly check that the environment is OK when running jobs, or specifically for the SQL extensions we need.

We could also potentially just add something to the manual's troubleshooting pages for the stock error.

Daniel-KM:
Yes the issue occurs for example when cli and apache have different versions of php.

When the fix is not set, the job stays in status "starting" and there is no log in common installs, so hard to find.

With the fix, the message is "An exception occurred in driver: could not find driver" . It comes from Doctrine exception : https://github.com/doctrine/dbal/blob/dc9b3c3c8592c935a6e590441f9abc0f9eba335b/lib/Doctrine/DBAL/DBALException.php#L169 "An exception occurred in driver:", that is not translated, and the exception comes from php itself, and is not translated too. 

Anyway, the fix displays now the message ("An exception occurred in driver: could not find driver") and a tip to fix it. The tip can be removed or based on the doctrine part of the message (normally this is the only possible message, else omeka won't work).

PR #2051, opened 2023-05-05
Show total in tab "Linked resources" even when there is only one page

Daniel-KM:
Some people want to get the total of linked resources, that is not displayed when there is only one page.


jimsafley:
Thanks for bringing this to our attention. To keep it consistent with other browse pages, we've decided to always show the pagination (075d35c7bffe3efb1a0e640b1795d5c44f44f545).

PR #2054, opened 2023-05-16
Added an event to get linked resources via subject values.

Daniel-KM:
Some people want to order subject values not by title, but by something else, for example by bibo:volume and bibo:issue. The feature is already implemented in module [Advanced Resource Template](https://gitlab.com/Daniel-KM/Omeka-S-module-AdvancedResourceTemplate), but an event is needed for that.


Daniel-KM:
Ok.

zerocrates:
Thanks.

Issue #2058, opened 2023-06-01
Resource page block default config for media doesn't include showing the media


zerocrates:
Default config for the media page main region is just `values`, and should include `mediaRender` as the first item instead.

This looks like it was an oversight that occurred when the `mediaRender` block was introduced in b951a29d39329c90c3a8f05a551c8c3467993757: the default render call was removed from the media show, but `mediaRender` wasn't added to the default media-page config.

A parallel issue exists in... probably all themes: we could cross-reference those to this issue to track updates for them.

PR #2059, opened 2023-06-01
Add media render to media resource pages

jimsafley:
#2058


zerocrates:
We'll want mediaRender to be on top of values. Matches the prior state of things pre-4.0.

PR #2067, opened 2023-06-27
Allow modules to add/modify API output formats

jimsafley:
Looking at [this post](https://forum.omeka.org/t/export-data-and-lido/18271) and it occurs to me that we don't have a way for modules to add output formats. We recently added the RDF outputs provided by EasyRdf, but it would be great to add an event or events where a module could transform the JSON-LD to, say, LIDO.

This PR adds two events that modules can use to customize the output:

- `api.output.formats`: Attach a handler to add a format/mime-type.
- `api.output.serialize`: Attach a handler to detect the format and set a custom output string

Implementations would look something like this:

```php
$sharedEventManager->attach(
    '*',
    'api.output.formats',
    function (ZendEvent $event) {
        $formats = $event->getParam('formats');
        $formats['xml'] = 'application/xml';
        $event->setParam('formats', $formats);
    }
);
$sharedEventManager->attach(
    '*',
    'api.output.serialize',
    function (ZendEvent $event) {
        $renderer = $event->getTarget();
        $model = $event->getParam('model');
        $format = $event->getParam('format');
        $payload = $event->getParam('payload');
        $output = $event->getParam('output');
        if ($format !== 'xml') {
            return;
        }
        $output = '<xml><foo>foobar</foo></xml>';
        $event->setParam('output', $output);
    }
);
```

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -1,14 +1,17 @@
 <?php
 namespace Omeka;
 
+use EasyRdf\Graph;
 use Omeka\Api\Adapter\FulltextSearchableInterface;
 use Omeka\Api\Representation\AbstractResourceEntityRepresentation;
+use Omeka\Api\Representation\RepresentationInterface;
 use Omeka\Entity\Item;
 use Omeka\Entity\Media;
 use Omeka\Module\AbstractModule;
 use Laminas\EventManager\Event as ZendEvent;
 use Laminas\EventManager\SharedEventManagerInterface;
 use Laminas\Form\Element;
+use Laminas\Json\Json;
 use Laminas\View\Renderer\PhpRenderer;
 
 /**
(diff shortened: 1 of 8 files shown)


application/Module.php
```diff
@@ -760,6 +775,99 @@ public function noindexItemSet(ZendEvent $event)
...
+
+        if ('jsonld' !== $format) {
+            return;
+        }
+
+        $eventManager = $this->getServiceLocator()->get('EventManager');
+
+        if ($payload instanceof RepresentationInterface) {
+            $args = $eventManager->prepareArgs(['jsonLd' => $output]);
+            $eventManager->trigger('rep.resource.json_output', $payload, $args);
```
jimsafley, line 797: I'm not sure we need this event anymore since this PR allows modules to modify the output.

application/Module.php
```diff
@@ -760,6 +775,99 @@ public function noindexItemSet(ZendEvent $event)
...
+            $output = Json::prettyPrint($output);
+        }
+
+        $jsonpCallback = (string) $model->getOption('callback');
+        if (!empty($jsonpCallback)) {
+            // Wrap the JSON in a JSONP callback. Normally this would be done
+            // via `$this->setJsonpCallback()` but we don't want to pass the
+            // wrapped string to `rep.resource.json_output` handlers.
+            $output = sprintf('%s(%s);', $jsonpCallback, $output);
+            $renderer->setHasJsonpCallback(true);
```
jimsafley, line 812: We set set `ApiJsonRenderer::$hasJsonpCallback` to true here and I don't know why. Any ideas?

zerocrates: It's used to have the content type get set to text/javascript, because JSONP is javascript code rather than JSON.

application/Module.php
zerocrates: Removing the event here would mean pushing this to a 5.0 so probably not justified.

PR #2069, opened 2023-06-29
Add events to allow modules to select values used in fulltext search

jajm:
Some values may not be wanted in the full-text index, either because of their content size (trying to index large amounts of text can seriously impact overall performances of the application) or because their content is irrelevant or add too much noise to the search results

This patch adds two new events:
- api.get_fulltext_text.value_criteria
- api.get_fulltext_text.value_annotation_criteria

Both events have `resource` and `criteria` as arguments and can modify `criteria`.
api.get_fulltext_text.value_annotation_criteria has an additional argument, `value`.

### DIFF
--- a/application/src/Api/Adapter/AbstractResourceEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractResourceEntityAdapter.php
@@ -694,7 +694,14 @@ abstract class AbstractResourceEntityAdapter extends AbstractEntityAdapter imple
         $services = $this->getServiceLocator();
         $dataTypes = $services->get('Omeka\DataTypeManager');
         $view = $services->get('ViewRenderer');
+        $eventManager = $this->getEventManager();
+
         $criteria = Criteria::create()->where(Criteria::expr()->eq('isPublic', true));
+        $args = $eventManager->prepareArgs(['resource' => $resource, 'criteria' => $criteria]);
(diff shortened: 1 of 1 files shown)


jajm:
My personal use case for this is to exclude extracttext:extracted_text from the fulltext search, because indexing large amount of texts takes too much time and can cause crashs (see related discussion: https://github.com/omeka-s-modules/ExtractText/pull/6)

jimsafley:
Thank you. It looks like a reasonable addition. We'll review this more thoroughly once we're able to confirm that the events will work with the use case.

jimsafley:
Getting back to this. Have you found this to work for your use case?

jajm:
Yes it works. Here's the code I used as an example:

```php
class Module extends AbstractModule
{
    public function attachListeners(SharedEventManagerInterface $sharedEventManager)
    {
        $sharedEventManager->attach('*', 'api.get_fulltext_text.value_criteria', [$this, 'onApiGetFulltextTextValueCriteria']);
    }

    public function onApiGetFulltextTextValueCriteria(Event $event)
    {
        $criteria = $event->getParam('criteria');

        $property = $this->getServiceLocator()->get('Omeka\EntityManager')->getRepository('Omeka\Entity\Property')->findOneBy(['localName' => 'extracted_text']);
        if ($property) {
            $criteria->andWhere(Criteria::expr()->neq('property', $property));
        }
    }
}
```

Issue #2062, opened 2023-06-12
Sort control for advanced search appears above module-added advanced search options


zerocrates:
We probably want sort either always on top or always on the bottom, or else outside the general area where inputs are added altogether.

"Always on the bottom" i think would mean we'd have to add it using an event.

PR #2075, opened 2023-07-10
Adjust sort control position

jimsafley:
See #2062


zerocrates:
Slightly skittish about just having sort on top always... but maybe it's actually fine.

Issue #2078, opened 2023-07-11
Favicon setting


zerocrates:
We could use the asset system to let people easily set a favicon. I don't think we need to worry about allowing for .ico itself at least at the moment, as png/gif/svg/etc work fine as favicons.

We could also let sites set different favicons... I'm not sure that's really warranted though.

PR #2079, opened 2023-07-11
Add global favicon setting

jimsafley:
See #2078


jimsafley:
I wonder if sites should use the favicon set in general settings if there is none set in site settings. It doesn't do that now.

PR #2084, opened 2023-07-18
Increase DB minimum versions

jimsafley:


Daniel-KM:
In order to allow json and spatial queries without issue in mysql and mariadb, it is recommended to use slightly higher versions (see #1969): 8.0.18 and 10.2.27. 
Because here the aim seems to include features available in various modules since a while, i had many issues with old versions of database during creation of the modules [Data Type Geometry](https://gitlab.com/Daniel-KM/Omeka-S-module-DataTypeGeometry) and [Annotate Cartography](https://gitlab.com/Daniel-KM/Omeka-S-module-Cartography) since 2018.

zerocrates:
So as I previously said on your PR, 10.2 and 5.7 are the closer match for each other in terms of features, so we're more likely to move to those together than to 10.2 and 8. 8 is also where MySQL is more divergent in terms of what it supports, so the value for us moving there is a little less... though if we did, we'd probably aim to require a later MariaDB as well to try to have the gap closed in terms of features some more. A close matchup with any MariaDB and MySQL 8 is harder than with 5.7, though.

Having a better base for the spatial functions in particular is a goal of this move, yes. MySQL 8 does add lots of interesting spatial functionality like SRIDs, but as far as I'm aware no version of MariaDB actaully supports these at the moment. The specific versions here were chosen for compatibility with each other.

JSON column support isn't an explicit goal of this move in particular but it is one of the other major things that gets changed/improved in the newer versions of both databases, and we of course do use a fair amount of JSON in various tables. However it has some similar types of problems, i.e. the JSON type is substantially different in MariaDB vs. MySQL: they do comparisons differently, store the data differently, support different functions and operators from each other. Even so, the 5.7.9 version specified here (the initial "final" release of 5.7) does have the JSON column type.

On the support front, MySQL 5.7 is still in its extended support period and had a new release as recently as June, but that period is coming to an end quite soon.

All that being said, if you do have specific experience and reasons for preferring these particular higher minimums from your work with your modules, we'd be happy to hear and consider them.

zerocrates:
@Daniel-KM We don't end up actually _needing_ higher versions for any of the geo features we're looking at for Mapping at the moment.

That being said, I'm not opposed to raising them anyway out of general principle for 4.1, but it would probably be only to the levels from this PR rather than the higher ones you've suggested, at least with the info I currently have.

I'll probably be making a decision on whether to do an increase for 4.1 relatively soon, so if you'd like to have any input please feel free. I'm assuming that, even in the absence of a yet-higher increase, you'd prefer this incremental increase over just maintaining the existing minimums.

PR #2092, opened 2023-08-29
Update README.md

katechap:


zerocrates:
I think you probably didn't mean to send this to us as a PR. Or at least not right now.

PR #2093, opened 2023-09-08
Vocabulary form improvements

jimsafley:


zerocrates:
I imagine that data attribute isn't going to actually get translated. Attributes except for a small handful that are pre-set don't get translated unless you explicitly set them to.

I believe this is an attribute not used elsewhere so we wouldn't have already set it up that way.

PR #2095, opened 2023-09-10
Add title to property and class select options

jimsafley:
It is possible for one vocabulary to have more than one distinct but identically labeled properties and classes. This ambiguity can cause [problems](https://forum.omeka.org/t/alternate-label-not-displaying-in-two-cases/19270/14). This PR adds a title attribute to vocabulary member select options, which contains the member comment, which is probably enough to disambiguate them. Maybe it should include the local name as well?


zerocrates:
Maybe _just_ the local name? That should be sufficient to disambiguate in the rare cases where a vocab decides to have matching labels for multiple terms, and is guaranteed to be there even if there aren't comments for every property.

Using the comment also seems likely to run us into issues of a desire to use alternate comments, as well as generally leads to larger tooltips... my instinct is to keep things short.

PR #2096, opened 2023-09-11
Performance/mariadb

Daniel-KM:
On bases with many linked resources, the search is very slow on mariadb. So this is a fix, just a composite index in fact, that allows to search instantly for the most common queris (https://github.com/omeka/omeka-s/blob/develop/application/src/Api/Adapter/AbstractResourceEntityAdapter.php#L281-L294).


Daniel-KM:
And the filling of the primary media fixes listing when items have many medias, but no primary media. I added a similar job in module [Easy Admin](https://gitlab.com/Daniel-KM/Omeka-S-module-EasyAdmin/-/blob/master/src/Job/DbItemPrimaryMedia.php?ref_type=heads). 

Ideally, the primary media should be filled automatically with the first media, but there are many places where it is not the case (in particular import modules). The other way is to use a stored sql procedure, but this is not very common in omeka.

zerocrates:
Just curious: have you seen this to not be the case on MySQL, or do you just have direct experience on it with MariaDB only?

zerocrates:
On the indexes: maybe wise to have is_public as the last component instead of the first? Admin users won't use is_public in their queries, and query suitability is usually based on a leftmost prefix of the indexed columns. So an index with is_public included last should work for both queries with and without an is_public query component.

Daniel-KM:
Yes, i tested on mysql (prod server on huma-num.fr) and mariadb (test server).

Daniel-KM:
For is_public, the order is important, because there is no index on is_public and when there is a composite index, the first index can be used as a single index (and the first with the second one, etc.). So the first thing is to have an index on it, so see the pr #2105 if you don't like composite index. It will resolve most of this issue too.

zerocrates:
Since we merged the independent is_public indexes you made the other PR for, I don't think we'll do this one, at least not at this time.

PR #2098, opened 2023-09-13
Report mail send failures more gracefully

### LINKED ISSUE #2097: Report mail send failure in a more friendly way
Instead of a "encountered an error" page we should probably just flash a warning when we try and fail to send an email, at least for user creation.

jimsafley:


zerocrates:
Only comment would be maybe we want to catch here more generically: \Laminas\Mail\Exception\ExceptionInterface should cover anything coming from the mail component.

Oh and: we should still log the exception even though we're flashing it instead of showing the error page.

PR #2105, opened 2023-10-20
Fix/index public

Daniel-KM:
For performance issue, there is the index on is_public, even if it is a boolean.


ivyrze (approved):

zerocrates:
Because this was based on the "consecutive or" branch that got rebased, it didn't apply cleanly as-is, so I just cherry-picked the actual change.

PR #2109, opened 2023-11-01
Added status API to api-local route.

Daniel-KM:
Is there a reason why api-local has no `__API__`?


zerocrates:
The `__API__` key is how the flip over to API key-based auth gets triggered. So if it gets set for api-local, it won't use the logged-in user's credentials as expected.

We can have an `__API_LOCAL__` and have isApiRequest in status also return true if that's set; I assume that status method is the issue here for you?

Issue #2116, opened 2023-11-21
HTML-type media is not text searchable


allanaaa:
The text contents of an HTML media aren't coming up in search results. Those media really should be full-text searchable.

PR #2117, opened 2023-11-22
Add way to get fulltext from media

jimsafley:
See #2116

### DIFF
--- a/application/src/Api/Adapter/MediaAdapter.php
+++ b/application/src/Api/Adapter/MediaAdapter.php
@@ -209,7 +209,7 @@ class MediaAdapter extends AbstractResourceEntityAdapter
             ->get($resource->getRenderer());
         $fulltextText = parent::getFulltextText($resource);
         if ($renderer instanceof FulltextSearchableInterface) {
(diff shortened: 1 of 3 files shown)


application/src/Api/Adapter/MediaAdapter.php
```diff
@@ -200,4 +201,16 @@ public function preprocessBatchUpdate(array $data, Request $request)
...
+    public function getFulltextText($resource)
+    {
+        $renderer = $this->getServiceLocator()
+            ->get('Omeka\Media\Renderer\Manager')
+            ->get($resource->getRenderer());
+        $mediaText = '';
+        if ($renderer instanceof FulltextSearchableInterface) {
+            $mediaText = $renderer->getFulltextText($this->getRepresentation($resource));
+        }
+        return sprintf('%s %s', parent::getFulltextText($resource), $mediaText);
```
zerocrates, line 214: This is really minor but: this introduces an extra space at the end of the fulltext when the renderer isn't fulltext-searchable or has no text.

PR #2118, opened 2023-11-27
Fix/deprecation warning

Daniel-KM:
Trim requires a string, not a null or integer.


zerocrates:
trim (and strlen, and similar functions) actually accepts integers without complaint; the deprecation change from several versions ago really only affected null.

Regardless, converting everything to strings and using strlen as the the filter does make what's happening here a little clearer probably.

PR #2120, opened 2023-11-27
Fix/translate aria label

Daniel-KM:


application/src/Service/Delegator/FormSelectDelegatorFactory.php
```diff
@@ -10,6 +10,8 @@ public function __invoke(ContainerInterface $container, $name,
         callable $callback, array $options = null
     ) {
         $formSelect = $callback();
+        $formSelect->addTranslatableAttribute('aria-label');
+        $formSelect->addTranslatableAttribute('title');
```
zerocrates, line 14: `title` is a default translatable attribute defined in the Laminas abstract form view helper so I don't believe this line is needed

application/src/View/Helper/SearchFilters.php
```diff
@@ -48,7 +48,7 @@ public function __invoke($partialName = null, array $query = null)
                 switch ($key) {
                     // Fulltext
                     case 'fulltext_search':
-                        $filterLabel = $translate('Search full-text');
+                        $filterLabel = $translate('Search full-text'); // @translate
```
zerocrates, line 51: `$translate` calls are detected by xgettext, so they don't need the `@translate` comment.

Generally the comment is only needed where the string literal is not used directly in a call to `$translate` or `$this->translate`.

zerocrates: Same comment applies to all the changes in this file, I believe

zerocrates (changes requested):
Mostly looks good to me, but I have a few comments: pretty much just a few unnecessary changes.

zerocrates:
I made the requested change on `title` myself, and also omitted the parts that made minor modifications to the block alignment selects (since those have since been moved/removed). For the redundant translate tags in the search filters, I just dropped that commit entirely.

PR #2121, opened 2023-11-28
Allow multiple file upload

jimsafley:


application/src/Media/Ingester/Upload.php
```diff
@@ -71,9 +71,16 @@ public function form(PhpRenderer $view, array $options = [])
...
+            'multiple' => true,
+            'data-info-template' => '
+            <div class="media-file-info">
+                <div class="media-file-thumbnail"></div>
+                <div class="media-file-size"></div>
+            </div>',
         ]);
         $field = $view->formRow($fileInput);
-        return $field . '<input type="hidden" name="o:media[__index__][file_index]" value="__index__">';
+        echo $field . '<input type="hidden" name="o:media[__index__][file_index]" value="__index__">';
```
zerocrates, line 84: What's going on here?

application/src/Media/Ingester/Upload.php
```diff
@@ -45,19 +45,19 @@ public function ingest(Media $media, Request $request, ErrorStore $errorStore)
         }
 
         $index = $data['file_index'];
-        if (!isset($fileData['file'][$index])) {
+        if (!isset($fileData['file'][$index][0])) {
```
zerocrates, line 48: Actually also, what's happening here? The indexing to zero seems like a problem... I would wager Laminas is undesirably adding the extra `[]` to the file inputs to require this extra indexing? We might need to look at being able to avoid that, rather than working around it on this end.

This isn't quite working as-is either: files are coming in with no source (so Untitled instead of having their filename) with this logged:

[06-Dec-2023 22:47:47 UTC] PHP Warning:  Undefined array key "name" in /var/www/html/omeka-s/application/src/Media/Ingester/Upload.php on line 60
[06-Dec-2023 22:47:47 UTC] PHP Warning:  Trying to access array offset on null in /var/www/html/omeka-s/application/src/Media/Ingester/Upload.php on line 60
[06-Dec-2023 22:47:47 UTC] PHP Warning:  Undefined array key "name" in /var/www/html/omeka-s/application/src/Media/Ingester/Upload.php on line 60
[06-Dec-2023 22:47:47 UTC] PHP Warning:  Trying to access array offset on null in /var/www/html/omeka-s/application/src/Media/Ingester/Upload.php on line 60

That looks like a one-off mistaken placement of the index when setting the source on line 60 so it would be a simple fix.

The more general problem is that this extra level of indexing is a change in the required API to upload files; it wouldn't be appropriate for 4.1 as-is. Coaxing Laminas into not messing with the name of the input (maybe just taking Laminas out of the rendering step) should avoid there needing to be any changes to the actual processing in the ingester.

PR #2126, opened 2023-11-28
Imported internal google fonts for self hosting.

Daniel-KM:
Google recommends to self host fonts and for privacy, it's better. And it allows to run Omeka fully locally or on an intranet without external connection (i have a site where omeka is on an intranet with another cms for public frontend and it's not a pretty sight).


zerocrates:
We're looking into this. My understanding is that we should be good with just woff/woff2: eot is only needed for ancient IEs we don't support anyway, svg similarly for ancient Safari versions, and ttf super old browsers generally.

PR #2127, opened 2023-11-28
Add title attribute to YouTube media iframe to improve accessibility

tomhigginsuom:
Adds a "title" attribute to the iframe for YouTube media to improve accessibility.

If no title was set on the media a default title is set instead.

Previous discussion on the forum:
* https://forum.omeka.org/t/youtube-media-iframe-elements-do-not-have-a-title/20198
* https://forum.omeka.org/t/how-do-i-add-alt-text-on-images-and-other-accessibility-features-for-screen-reader-users-when-the-alt-text-module-does-not-appear-to-do-that/12719


tomhigginsuom:
Implemented by https://github.com/omeka/omeka-s/pull/2314

PR #2147, opened 2024-02-13
New block layout sidebar

jimsafley:


application/view/omeka/site-admin/page/edit.phtml
```diff
@@ -108,12 +108,16 @@ $form->prepare();
     <div id="layout-preview"></div>
 </div>
 
-<div id="block-layout-data-sidebar" class="sidebar" data-api-endpoint-url="<?php echo $escape($this->url('api')); ?>">
-    <?php 
-        echo $this->partial('common/block-layout-sidebar.phtml', [
-            'blockLayoutDataForm' => $blockLayoutDataForm
-        ]); 
-    ?>
+<?php $collapsibleTemplate = sprintf('<a href="#" class="expand" aria-label="%1$s" title="%1$s"></a><div class="collapsible"></div>', $translate('Expand')); ?>
```
zerocrates, line 111: This template and the related data attribute  on the div below are vestiges of the JS-based system that we're no longer using, I believe.

PR #2152, opened 2024-02-19
Separate record and text in full text search

Daniel-KM:
The first commit add an index on is_public, that it was missing in last pr. This is important, because all queries use it, so it speed end users queries a lot.

The second commit separate the fulltext field in a field "record" (for the record) and a field "text" (for the text: transcription, ocr, content, etc.). Many librarians make a distinction between to search the record and to search the content and this hard to manage without recreate another search table. This second commit is complete, but requires some more work, in particular a setting to determine what is full text (bibo:content or extracttext:extracted_text and some other) and another optional setting to let user select full text or not, by default or not. To use the event is possible, but complex.

So you can cherry-pick the first and say me what you think about the second.

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -710,7 +710,7 @@ class Module extends AbstractModule
         }
         $qb = $event->getParam('queryBuilder');
 
-        $match = 'MATCH(omeka_fulltext_search.title, omeka_fulltext_search.text) AGAINST (:omeka_fulltext_search)';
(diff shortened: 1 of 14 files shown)


zerocrates:
Are you seeing big improvements from the is_public index?

My understanding has been that there's really no support for combining the usage of a regular index with a fulltext one, so I'm somewhat surprised there'd be a large improvement there. I'd assume the fulltext would usually be the more specific index and get used.

PR #2159, opened 2024-03-07
Fixed missing use statement in FormCollectionElementGroups.

Daniel-KM:


zerocrates:
Have you tried this out with the fix and a "grouped" element that's in a fieldset? It's definitely a bug, but I have a suspicion that supporting elements that are in fieldsets wouldn't work very well anyway: it seems to me that it would probably render the element twice: once in the group and once in the fieldset.

We might instead want to just remove the code that tries to walk down into fieldsets entirely, rather than fixing the use.

Daniel-KM:
This is a bug i see when trying module bulk edit, that adds many fieldsets to the batch form.

Daniel-KM:
Yes, the group of omeka matches the standard fieldset of html and laminas, but without adding a level in names. So for now, i fixed the new behavior in my modules to avoid to remove fieldsets.

zerocrates:
OK, and I've just gone ahead and tested with a simple fieldset and it's definitely preferable to have it correctly descend into the fieldsets in groupElements, otherwise we're just getting completely broken mis-rendering of the fieldset as an "element" once there are defined groups. So this is definitely a straightforward improvement.

PR #2167, opened 2024-03-14
Filter default sites for items user setting

### LINKED ISSUE #2166: User's item default settings don't work
When a user has a default item template, item set(s), and site(s) for all their new items, the resource template seems to work (and automatically adds the associated class), but the default item set and default site don't automatically get set.

jimsafley:
This is an attempt to resolve a conflict between the "assign_new_items" site setting and the "can-assign-items" user permission. Put simply, a site can be set to automatically assign itself to new items, but the user can lack the permission to assign items to the site. This conflict could lead to a broken "Site" tab on the item add page (see #2166).

This PR fixes the conflict by a) limiting the sites on the site selector sidebar to those that the current user has "can-assign-items" permission; and b) adding the sites that are "assign_new_items" but restricted by "can-assign-items" to the "Site" tab, giving users the opportunity to assign these restricted sites, but _only_ on the item add page. A few other changes to the site selector element, the item adapter, and JS were necessary to accomplish this.

### DIFF
--- a/application/src/Api/Adapter/ItemAdapter.php
+++ b/application/src/Api/Adapter/ItemAdapter.php
@@ -258,13 +258,11 @@ class ItemAdapter extends AbstractResourceEntityAdapter
                 if (!$site) {
                     // Assign site that was not already assigned.
                     $site = $siteAdapter->findEntity($siteId);
-                    if ($acl->userIsAllowed($site, 'can-assign-items')) {
+                    if ($acl->userIsAllowed($site, 'can-assign-items') || ($isCreate && $assignNewItemsSites->contains($site))) {
                         // A user with the "can-assign-items" privilege can assign
-                        // a site at any time.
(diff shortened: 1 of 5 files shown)


application/src/Api/Adapter/ItemAdapter.php
```diff
@@ -255,6 +259,12 @@ public function hydrate(Request $request, EntityInterface $entity,
                     // Assign site that was not already assigned.
                     $site = $siteAdapter->findEntity($siteId);
                     if ($acl->userIsAllowed($site, 'can-assign-items')) {
+                        // A user with the "can-assign-items" privilege can assign
+                        // a site at any time.
+                        $sites->set($site->getId(), $site);
+                    } elseif ($isCreate && $assignNewItemsSites->contains($site)) {
```
zerocrates, line 265: This is probably clearer written as an or in the initial if statement

PR #2173, opened 2024-04-04
Fix linked resources interface

jimsafley:
This is an attempt to fix the linked resources resource page block, specifically the redundant select options and inconsistency between the select menu and the results.

### DIFF
--- a/application/src/Api/Adapter/AbstractResourceEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractResourceEntityAdapter.php
@@ -676,16 +676,29 @@ abstract class AbstractResourceEntityAdapter extends AbstractEntityAdapter imple
         foreach ($qb->getQuery()->getResult() as $result) {
             if ($result['property_alternate_label']) {
                 $label = $result['property_alternate_label'];
+                $labelIsTranslatable = false;
                 $resourceTemplatePropertyId = $result['resource_template_property_id'];
             } elseif ($result['resource_template_property_id']) {
                 $label = $result['property_label'];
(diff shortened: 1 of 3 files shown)


application/src/Api/Adapter/AbstractResourceEntityAdapter.php
```diff
@@ -676,16 +676,29 @@ public function getSubjectValueProperties(Resource $resource, $resourceType = nu
...
                 $label = $result['property_label'];
+                $labelIsTranslatable = true;
                 $resourceTemplatePropertyId = 0;
             }
             $results[$result['property_id']][$label]['resource_template_property_ids'][] = $resourceTemplatePropertyId;
             $results[$result['property_id']][$label]['term'] = $result['term'];
+            // The shared label is translatable if at least one of the individual
+            // labels is a property label. A shared label is not translatable if
+            // all the individual labels are alternate labels.
+            if (isset($results[$result['property_id']][$label]['label_is_translatable'])) {
```
zerocrates, line 695: I think it might be a little clearer/simpler to do it like this:

```
if ($labelIsTranslatable) {
   $results[$result['property_id']][$label]['label_is_translatable'] = true;
}
```

to replace all of what's up here, and then down when building $subjectValueProperties below do

```
'label_is_translatable' => $data['label_is_translatable'] ?? false,
```

So you only ever set label_is_translatable to the $results if it's true, then if it's missing (so it was never true for that shared label) you just mark it false when returning. And then you don't have to be checking if the flag is already set for the current label every time through the loop.

Issue #2161, opened 2024-03-07
oEmbed block


allanaaa:
Is there any way to change the oEmbed URL after you first save the page? The box is appearing greyed-out. It's inconvenient if you wanted to insert a different video into the same spot on the page, especially after seeing what kind of layout constraints are being sent by the source. 

<img width="810" alt="Screenshot 2024-03-07 at 13 06 08" src="https://github.com/omeka/omeka-s/assets/28012148/ec0ecd6c-6aeb-4faf-ace8-1f253b7a4d24">

jimsafley:
The oEmbed URL is readonly because there's no way to tell that it's been changed. The only way to make this work is to fetch the oEmbed response every time the user saves the page.

zerocrates:
I think we'll deal with this post-4.1.

zerocrates:
This is done.

PR #2177, opened 2024-04-08
Allow updating of URL in oEmbed block layout

jimsafley:
See #2161


zerocrates:
I don't love that we have to send out the current URL as old_url when we should just be able to compare with the data before it's updated, but I understand that the way we have onHydrate working doesn't really allow for that.

zerocrates:
We might want to look and see if we can do something in the block saving code to allow the kind of "compare to what's already saved" workflow I alluded to in my previous comment. Maybe an extra method that would be called before the new `data` gets set, or something? But I don't think that needs to be a blocker on this.

PR #2179, opened 2024-04-09
Set lazy loading for images.

Daniel-KM:
Hi,
Some people load assets that are very heavy, so lazy loading is now supported widely, so the attribute can be set on assets and in fact in any image. It is particularly useful on scanned books with many files.


Daniel-KM:
https://caniuse.com/loading-lazy-attr

PR #2187, opened 2024-04-18
unset prop value in api

mquantin:
This light and simple api method allows to unset a value (property) in an item representation.
- Because it's better than filter it xx times in loops
- Because values of item representation are protected
- Because the view file can't do such filtering (with parameters)
- Because it does not affect the item itself, so it is safe

thx

### DIFF
--- a/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
+++ b/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
@@ -388,6 +388,16 @@ abstract class AbstractResourceEntityRepresentation extends AbstractEntityRepres
     }
 
     /**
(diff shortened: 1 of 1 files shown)


zerocrates:
Where are you'd thinking you'd use this, in a theme, something like that?

mquantin:
yes exactly ! in a theme. dedicated to CIDOC CRM (event oriented data modeling)
I display side-by-side the object (1 omeka resource) and events (1 omeka resource for each) of its life-cycle.
As the object and its life-cycle events are on the same page, it is useless to display the property saying "this event concerns this object". It is implicit. 
The easiest and cleanest way is to modify the item representation.

PR #2196, opened 2024-05-24
Fix missing check if original folder exits.

andreaswarnaar:
This pull request fix missing check if original folder exits. 

In early stages when no files are handled, the ```original``` folder does not exist yet, with the result of an error when calling ```disk_free_space ``` with a non existing folder path.


zerocrates:
The test failures here are about code style: the code style requires a space between `if` and the condition.

andreaswarnaar:
ai, missed that one. 
I updated the PR

PR #2200, opened 2024-06-07
Implemented module page template and module block template

Daniel-KM:
The possibility to create multiple templates for blocks is very practical. There is just a last point: let modules create templates too. For example a module can implement a specific template for browse preview, whatever the theme is.

This feature is implemented in the new version of the module BlockPlus, but it should be usable by any module.

Technically, i prefered to inject the module templates (a list of key-value pairs for path/label) inside the theme data because it is the simplest way, but it can be done in a heavy way.

### DIFF
--- a/application/config/module.config.php
+++ b/application/config/module.config.php
@@ -746,6 +746,7 @@ return [
             ],
         ],
     ],
+    'page_templates' => [],
     'block_templates' => [],
     'block_layouts' => [
         'invokables' => [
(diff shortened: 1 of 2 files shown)


zerocrates:
Thanks. Letting modules provide these across all themes seems like a fine idea.

I did merge this, though one thing does occur to me now: you're merging the module templates into the theme's config... should it be the other way around, just a flipping of the arguments to array_merge_recursive? We often have themes be the highest-priority override.

PR #2205, opened 2024-06-13
Add a GLB (3D) file renderer

jimsafley:


erseco:
Thanks again for this PR — it was a great starting point!

We ended up implementing a full 3D model viewer module for Omeka S, inspired in part by the code here. Our version supports both **GLB** and **STL** files using `model-viewer` and `three.js` and includes some additional features like background customization, auto-rotation, and grid display.

You can find it here: [https://github.com/ateeducacion/omeka-s-ThreeDViewer](https://github.com/ateeducacion/omeka-s-ThreeDViewer)

We hope it can complement or inform further development of 3D support in core Omeka S. Thanks again for the initiative!

PR #2224, opened 2024-09-04
Fixed searching full text with scalar results.

Daniel-KM:


zerocrates:
What's the bug here, or a reproducer for it? It's not as simple as just doing returnScalar while searching fulltext, I guess, because I tried that without issue.

zerocrates:
Probably also worth asking the database vendor, version, and SQL mode... ONLY_FULL_GROUP_BY is standard in MySQL 8+ but they have functional dependency support, while Maria doesn't know about functional dependencies but ONLY_FULL_GROUP_BY is not a default mode there.

I'm not sure supporting ONLY_FULL_GROUP_BY on Maria, if that's the configuration being used here, is a goal.

PR #2236, opened 2024-11-04
Fixed missing event manager for form AssetEditForm.

Daniel-KM:


zerocrates:
The bug here is that the event manager isn't being injected for this form so you're getting a "local only" one that doesn't connect to the shared event manager and modules therefore can't attach listeners to it?

Daniel-KM:
The form is not declared in the config so it does not load the shared event manager. In fact, the trait allows to load one event manager, but there is no events inside it. So it is like the SiteFormFactory for example.

zerocrates:
Since so many forms need this injected, I think I might instead fix this with an initializer that auto-injects the event manager for any form that implements EventManagerAwareInterface: that pattern's already used by Laminas itself for the top-level service manager and controllers.

zerocrates:
I went ahead with that strategy, and my review of the forms also noted the same problem in the newer BlockLayoutDataForm and PageLayoutDataForm.

PR #2241, opened 2024-11-27
Simplified management of roles and allowed to made them inheritable.

Daniel-KM:


zerocrates:
So the idea here is that the value for each role in that "roles" key is the roles it inherits from, and the default ones just don't inherit and that's why they're all empty arrays?

I feel like I'd prefer to see these three config arrays squashed into one "roles" key with the structure

```
'role_id' => [
    'label' => /* string */,
    'admin' => /* bool */,
    'parents' => /* string[] */.
]
```

and then we don't have to repeat the ids several places, plus an explicit "parents" subkey feels a little more self-documenting.

Daniel-KM:
I have no preference for the structure of the config, i recopied the old one of \Omeka\Permissions\Acl that has an array for roles and another one for labels. 

The main point is to be able to manage roles and new roles simpler, for example guest and derivated ones. Currently, it is complex, because the roles should be created early to be fully managed in previous modules.

For parents, this is a native feature of Laminas Acl (https://docs.laminas.dev/laminas-permissions-acl/usage/#multiple-inheritance-among-roles) and the list of parents just need to be passed [when the role is created](https://github.com/omeka/omeka-s/pull/2241/files#diff-9c68a4f61d08b948e16a0be6ebd6ff6d88807db141938bd2fce5c498908db742R70-R74). The default roles don't use parents, so they are empty array, but a module can add new rights for them just by adding one or more new roles.

Daniel-KM:
So at your choice.

PR #2245, opened 2024-12-09
Add property is/is not resource queries to advanced search

jimsafley:
Adds the "is resource" (re) and "is not resource" (nre) search types to the advanced search.


jimsafley:
Closing in favor of #2246

PR #2247, opened 2024-12-11
Add IIIF block layouts

jimsafley:
This PR adds the "IIIF Presentation" and "IIIF image" block layouts. They operate similarly to the IIIF media options.


jimsafley:
@zerocrates Testing is complete (#2380). Ready to merge.

Issue #2265, opened 2025-02-28
Add API request option for search to skip counting total results


zerocrates:
Sometimes we're not interested in the total count for a request (we're getting the full results, or we just aren't going to look at the count at all), and getting the count runs another potentially expensive query.

We could introduce a request option to skip counting, and either return null for the count, or maybe better, just use PHP's `count()` to get the count of whatever the current request returned, if that option is set. Optimistically you could cut API response time in half for ones that skip the count, though the reality would be less. This would basically be the opposite of the "just get the count" feature we have when you pass a zero limit in the query.

Places we could use this would include the queries used to get options for `select` elements and other UIs that show all the options at once, probably the `searchOne` API helper method, and probably others.

PR #2266, opened 2025-03-01
Add option to skip Paginator count

jimsafley:
See #2265

This adds an API request option to skip the Paginator count, but instead of passing `skipPaginatorCount=true` to every `search()` that needs it, I've set a heuristic that automatically sets the option to true when none of the queries used for pagination are set. I can't think of a situation where there's no pagination, but there's still a need for a total count calculated by the Paginator.

### DIFF
--- a/application/src/Api/Adapter/AbstractEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractEntityAdapter.php
@@ -273,6 +273,19 @@ abstract class AbstractEntityAdapter extends AbstractAdapter implements EntityAd
         // Add the LIMIT clause.
         $this->limitQuery($qb, $query);
 
+        // Set whether to skip the Paginator count. Setting to true, or setting
+        // none of the queries used for pagination, will skip getting the total
+        // count from the Paginaor.
+        $skipPaginatorCount = $request->getOption('skipPaginatorCount');
+        if (null === $skipPaginatorCount
+            && !isset($query['page'])
+            && !isset($query['per_page'])
+            && !isset($query['limit'])
+            && !isset($query['offset'])
+        ) {
+            $skipPaginatorCount = true;
+        }
+
         // Before adding the ORDER BY clause, set a paginator responsible for
         // getting the total count. This optimization excludes the ORDER BY
         // clause from the count query, greatly speeding up response time.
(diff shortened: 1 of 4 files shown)


zerocrates:
I'm thinking reverse the direction of the option, to follow the example of other ones we added like `finalize`/`initialize`/`flushEntityManager`: default it to true and the user can set it as false, so `countQuery` or whatever.

You could actually write it so the default passed to getOption is just the result of evaluating your fallback checks:

```
$countQueryDefault = isset($query['page']) || isset($query['limit']) || isset($query['offset']);
```

which should handle the "automatic" aspect as well as letting you override in either direction, with no extra checks. I'd also drop per_page to match how the pagination works: `per_page` has no effect without `page`. (You could maybe instead look at the `$qb` and call `getMaxResults`/`getFirstResult` to keep the checks on the pagination query args in only one place... but I'm not sure that's very important here).

zerocrates:
Hmmm actually we use `getMaxResults` already to do the limit skip stuff, so maybe I'm leaning that we should actually use those for this and then we have all the pagination/limiting handled only in limitQuery.

```
$this->limitQuery($qb, $query);

$maxResults = $qb->getMaxResults();
$firstResult = $qb->getFirstResult();
$countQueryDefault = $maxResults || $firstResult > 0;
```

Then down at the bottom around the call to the real paginator to get the results, we'd replace the current

```
if ($qb->getMaxResults() || null === $qb->getMaxResults()) {
```

with

```
if ($maxResults !== 0) {
```

application/src/Api/Adapter/AbstractEntityAdapter.php
```diff
@@ -272,13 +272,21 @@ public function search(Request $request)
...
         // Before adding the ORDER BY clause, set a paginator responsible for
-        // getting the total count. This optimization excludes the ORDER BY
-        // clause from the count query, greatly speeding up response time.
-        $countQb = clone $qb;
-        $countQb->select('1')->resetDQLPart('orderBy');
-        $countPaginator = new Paginator($countQb, false);
+        // getting the total count (unless configured not to or when no pagination
+        // is set.
+        $countQueryDefault = $maxResults !== null || $firstResult > 0;
+        $countQuery = $request->getOption('countQuery') ?? $countQueryDefault;
```
zerocrates, line 282: getOption takes as its second arg the default value if I'm remembering correctly... should be the same behavior as this using the `??` but I think we'd want to keep to that established pattern

zerocrates:
The API view helper's searchOne is another case where we should probably turn this on by default... it won't get the default triggers here since it sets a limit of 1 but it's typically (always?) getting used in a context where the caller is just trying to get the content and doesn't care about the count. And by the same token the controller plugin's searchOne.

PR #2270, opened 2025-03-10
Removed zend framework bridge.

Daniel-KM:
This library is useless now, no module or dependencies uses it.

### DIFF
--- a/application/data/overrides/AbstractProxyFactory.php
+++ b/application/data/overrides/AbstractProxyFactory.php
@@ -15,6 +15,8 @@ use function in_array;
 
 /**
  * Abstract factory for proxy objects.
+ *
+ * @deprecated The AbstractProxyFactory class is deprecated since doctrine/common 3.5.
  */
 abstract class AbstractProxyFactory
 {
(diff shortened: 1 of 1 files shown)


Daniel-KM:
There is an issue but "it works on my computer", so some investigation to do.

Daniel-KM:
So it requires to updated doctrine orm config and to replace deprecated doctrine cache by symfony cache.

And in fact, i see that there are some other abandoned packages that should be replaced by symfony or psr equivalent:  

- Package laminas/laminas-dom is abandoned, you should avoid using it. Use symfony/dom-crawler instead.
- Package laminas/laminas-json is abandoned, you should avoid using it. No replacement was suggested.
- Package laminas/laminas-loader is abandoned, you should avoid using it. No replacement was suggested.
- Package laminas/laminas-log is abandoned, you should avoid using it. Use monolog/monolog instead.
- Package laminas/laminas-mail is abandoned, you should avoid using it. Use symfony/mailer instead.
- Package laminas/laminas-math is abandoned, you should avoid using it. No replacement was suggested.
- Package laminas/laminas-mime is abandoned, you should avoid using it. Use symfony/mime instead.
- Package lsolesen/pel is abandoned, you should avoid using it. No replacement was suggested.

Daniel-KM:
More changes are needed to use entities annotations.

Daniel-KM:
So to use the symfony cache requires to update all entities to add a use statement and @ORM or php 8 attributes (annotations are deprecated). But it breaks modules with entities. So this pr can be used for next version and for a more complete upgrade (see #2272).

PR #2276, opened 2025-03-12
Used api-local for local use in admin interface.

Daniel-KM:
In some cases, the api should be disabled, but to disable it breaks admin interface. So replacing the url api by api-local fix it. This is the purpose of this route anyway.


zerocrates:
Looks good to me; agreed that this is pretty much what api-local is for.

PR #2291, opened 2025-04-12
Add events on setting insert, update, and delete

jimsafley:
Settings bypass the API and ORM during insert, update, and delete, so there is no way to detect changes. This PR adds events during these operations so handlers can detect them.


zerocrates:
Only thing I'd change here is, move the "implements" for the EventManagerAwareInterface to the AbstractSettings class. I don't think we need to declare that all things implementing our settings interface need to be event manager aware.

PR #2314, opened 2025-06-27
Make YouTube iframes accessibility compliant

Trott:
The iframes for embedding YouTube videos should have an accessible name even before the video loads. This sets the title for the iframe element. It's entirely possible, likely, and acceptable that the YouTube embed JavaScript will overwrite it after the video loads.

### DIFF
--- a/application/src/Media/Renderer/Youtube.php
+++ b/application/src/Media/Renderer/Youtube.php
@@ -37,7 +37,7 @@ class Youtube implements RendererInterface
         }
         $url->setQuery($query);
         $embed = sprintf(
-            '<iframe width="%s" height="%s" src="%s" title="%s" frameborder="0" %s></iframe>',
+            '<iframe width="%s" height="%s" src="%s" title="%s" frameborder="0"%s></iframe>',
             $view->escapeHtml($options['width']),
             $view->escapeHtml($options['height']),
             $view->escapeHtml($url),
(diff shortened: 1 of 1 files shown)


Trott:
I might have been looking in the wrong place, but I didn't see any tests that involve the YouTube embedding code. If there are tests for this functionality, I'm happy to modify them to add a check for the `title` attribute.

Trott:
Testing locally, this doesn't quite work as the title is never in the object, only the URL. Putting in draft mode while I fix it...

Trott:
OK, now it works, including the fallback. Ready for review.

Trott:
> the space before that last `%s` is purposely omitted. The space there, if needed, gets built into the `allowfullscreen` that gets added on there.

OK, fixed. Thanks!

Trott:
Is there anything I can do to move this forward?

Trott:
Thanks for merging! Is there a timeframe for releases? It looks like it's been more than a year since a release and I would like to grab a release with this in it. (Maybe there are nightlies or something like that? I guess I could my hosting provider to install via `git checkout` but I'd rather not and they'd likely say no anyway.)

zerocrates:
I don't have a specific time for you, but we're getting closer to doing a release.

There are no nightlies; your options are either the releases or using a Git branch. You could always patch your current install if you need the change now.

PR #2317, opened 2025-07-07
Full json serializing representations

Daniel-KM:
There was a long time issue "doctrine new entity" and this pr fix it. 

Furthermore, it fixes most of the overflow of the memory, in particular for big import. It works for a migration of 300000 items from an old system via oai-pmh.

### DIFF
--- a/application/src/Api/Representation/AbstractRepresentation.php
+++ b/application/src/Api/Representation/AbstractRepresentation.php
@@ -48,10 +48,14 @@ abstract class AbstractRepresentation implements RepresentationInterface
     /**
      * Get a JSON serializable instance of DateTime.
      *
-     * @param \DateTime $dateTime
+     * DateTime may be null, in which case json-serialized output will be null.
+     * To get json-serializable output of current DateTime, pass `new \DateTime()`
+     * as parameter.
+     *
+     * @param \DateTime|null $dateTime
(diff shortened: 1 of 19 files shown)


Daniel-KM:
The point is that the old serialization didn't serialize objects, in particular owner and items (with values, properties etc), so in a bulk loop, new ones were fetched so there were two owner objects, etc. in the doctrine cache, then another new one in next loop, etc., whatever the use of entity manager flush or clear.

Daniel-KM:
Yes, json_encode() can be used anywhere, but i used jsonSerialize() for consistency with existing code and because json_encode() is calling jsonSerialize(), so it is useless when called inside the object. But they are the same (they should be).

The aim is to remove all representations early, because they store the resource entity and related resource entities. Even if it is just a link to a php object, to keep representations between loops implies some objects are never cleared from memory in some context. 

I didn't check for memory precisely. Usually, for big migrations, i need to increase the memory a lot, and with such a change, i don't need it. Maybe it is related to some other improvements done anywhere else in modules that fix the "new entity" issue at the same time and in that case the pr is not needed for this reason. 

The point is that `$resource->jsonSerialize()` should return a json-ld representation anywhere, not a simple json with php objects inside it, and not only via the api endpoint.

PR #2319, opened 2025-07-09
Omeka query builder

jimsafley:
Adds new, preferred methods to create named parameters and aliases used by the query builder, and deprecates the old methods. This change eliminates the need to reset the param/alias index within the adapter. New query builders "reset" the index automatically for free. See https://github.com/omeka/omeka-s/pull/2244#issuecomment-3050638880 for some background on why the indexes are important.

| Deprecated method | New method |
| ---- | ---- |
| `\Omeka\Api\Adapter\AbstractEntityAdapter::createNamedParameter()` | `\Omeka\Db\QueryBuilder::createNamedParameter()` |
| `\Omeka\Api\Adapter\AbstractEntityAdapter::createAlias()` | `\Omeka\Db\QueryBuilder::createAlias()` |

Modules using the deprecated methods should switch when able. For example, in module adapters, change these:

```php
$namedParam = $this->createNamedParameter($qb, $value);
$alias = $this->createAlias();
```

To these:

```php
$namedParam = $qb->createNamedParameter($value);
$alias = $qb->createAlias();
```

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -453,8 +453,9 @@ class Module extends AbstractModule
         }
 
         $adapter = $event->getTarget();
-        $itemAlias = $adapter->createAlias();
+
         $qb = $event->getParam('queryBuilder');
+        $itemAlias = $qb->createAlias();
         $qb->innerJoin('omeka_root.item', $itemAlias);
 
         // Users can view media they do not own that belong to public items.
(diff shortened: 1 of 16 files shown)


Daniel-KM:
See #2321 too: add optional type to method createNamedParameter().

PR #2321, opened 2025-07-15
Based on #2319, add the type to createNamedparameter()

Daniel-KM:


zerocrates:
That's a good improvement and this is a good time to take care of it. I cherry-picked the commit.

Issue #2293, opened 2025-04-16
Hidden sites showing up on cross site search to not logged in users


C-Norton:
When search type "All Sites" is selected, the sites listed include links to sites the current user, logged in or not, do not have permission to view. This, of course, results in a dead link. Furthermore, test sites and sites under creation that are still marked as private are visible.

Ideally, this list should be filtered to results that the user has permission to view.

C-Norton:
I've made appropriate changes in 

https://github.com/C-Norton/omeka-s/tree/develop

changed 
application/view/common/cross-site-search/results.phtml and application/view/common/cross-site-search/item-results.phtml

Unfortunately, I'm having trouble running the unit tests, so can't do anything more than manual testing, but it appears to work. I've indicated the areas of change with comments. I'd like to work with the team to get this to the point where it can be committed back into the project
@zerocrates

PR #2341, opened 2025-08-07
 Filtered visible sites on cross site search to only sites the user has permission to view #

C-Norton:
Added appropriate unit tests (hopefully in the right directory). This aligns with [issue 2293](https://github.com/omeka/omeka-s/issues/2293). Made this change for item sets as well


C-Norton:
Hi, just looking for an update here, as this is a concern for our institution, and we don't want to run a modified version of Omeka, vs one that's been verified.

zerocrates:
OK, a couple comments:

- The indentation changes on item-results.phtml aren't necessary so I'd like to lose those.
- However, the fact that you're not changing item-results.phtml at all in a meaningful way is odd given that you're changing all the other results pages; this is probably an oversight.
- More broadly, I think your method of detecting which sites are visible that's getting used on all the results pages is overly complicated... my instinct would be to just call `if ($site->userIsAllowed('read')):` to check these. Probably as simply as just doing it as a single line in the existing loops used to display the sites for each result. That's probably fine on its own... we could memoize those results if performance was a concern... but those permission checks should be quite fast.

zerocrates:
To be clear about my last point, I put the simpler version I'm talking about there up as a branch, cross-site-visibility.

Issue #2349, opened 2025-09-09
Admin item set show doesn't list related sites


kimisgold:
In the admin interface, the item show view details which sites it belongs to. The item set show view should do so as well.

PR #2352, opened 2025-09-11
List sites in item set details sidebar

jimsafley:
Per #2349, this PR lists sites in the item set details sidebar


zerocrates:
We took #2346 instead of this one.

PR #2359, opened 2025-09-18
Fix garbled first character of multibyte usernames in "User permissions" sidebar.

rish314:
# Summary
Fixed an issue where the first character of multibyte usernames was garbled in the right sidebar of the "User permissions" page.

Before:
スクリーンショット 2025-09-19 0 04 12

After:
スクリーンショット 2025-09-19 0 00 17

### DIFF
--- a/application/src/View/Helper/UserSelector.php
+++ b/application/src/View/Helper/UserSelector.php
@@ -43,8 +43,8 @@ class UserSelector extends AbstractHelper
     {
         $usersByInitial = [];
         foreach ($users as $user) {
-            $initial = mb_substr($user->name(), 0, 1);
-            $upper_initial = mb_strtoupper($initial);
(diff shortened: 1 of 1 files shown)


zerocrates:
Ah sure, indexing into the string is splitting apart multibyte UTF-8 chars if they're first.

Happy with the fix. Two things:
1) you'll see that you've tripped the code style checker; it's just complaining about the formatting of the stub classes in your test, so if you could fix those
2) I'd rather not have things depending on the mbstring config settings, so please pass the `UTF-8` argument at the end of your `mb_*` calls

rish314:
@zerocrates 
Thank you for your comments. I’ve committed the changes based on your advice. Please check them, and let me know if you notice anything else.

zerocrates:
Looks good. Thanks for the contribution.

Issue #2391, opened 2025-12-09
New displayValues() option properties only renders in default order


jlayt:
The new displayValues() option 'properties' to select which values to render is very welcome, however the implementation uses the default order of the values to output rather than the passed in selection order. This would be a useful option to have, perhaps 'propertyOrder' or 'orderedProperties'?

https://github.com/omeka/omeka-s/blob/4bb3350dad55fc50ce7b4370bec978ba979b968e/application/src/Api/Representation/AbstractResourceEntityRepresentation.php#L491

zerocrates:
Yeah, a "propertiesRespectOrder" boolean or something like that? I don't think that would be a problem.... we'd just have to swap around some of the array intersecting/merging here to get that result.

PR #2393, opened 2025-12-13
Add displayValues option to respect property order

jimsafley:
See #2391. Adds "propertiesRespectOrder" option to `AbstractResourceEntityRepresentation::displayValues()`. Setting it to `true` changes default behavior to respect the order of properties passed in the "properties" option. The default behavior is to maintain the original property order, irrespective of the passed order. For example, this:

```php
$item->displayValues(['properties' => ['dcterms:subject', 'dcterms:description', 'dcterms:title']]);
```

will, in most cases, output "Title, Description, Subject" because that's the order defined on the resource itself. Whereas, this:

```php
$item->displayValues(['properties' => ['dcterms:subject', 'dcterms:description', 'dcterms:title'], 'propertiesRespectOrder' => true])
```

will output "Subject, Description, Title" because "propertiesRespectOrder" is `true`, and that's the order passed in the "properties" option.

### DIFF
--- a/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
+++ b/application/src/Api/Representation/AbstractResourceEntityRepresentation.php
@@ -481,6 +481,7 @@ abstract class AbstractResourceEntityRepresentation extends AbstractEntityRepres
         $options['viewName'] ??= 'common/resource-values';
         $options['siteId'] ??= null;
         $options['properties'] ??= [];
+        $options['propertiesRespectOrder'] ??= false;
         $options['excludeProperties'] ??= [];
 
         $services = $this->getServiceLocator();
@@ -488,7 +489,11 @@ abstract class AbstractResourceEntityRepresentation extends AbstractEntityRepres
 
         // Filter values by the "properties" and "excludeProperties" options.
         if ($options['properties']) {
-            $values = array_intersect_key($values, array_flip($options['properties']));
+            $termsAsKeys = array_flip($options['properties']);
+            $values = $options['propertiesRespectOrder']
+                ? array_replace($termsAsKeys, $values)
+                : $values;
+            $values = array_intersect_key($values, $termsAsKeys);
         }
         if ($options['excludeProperties']) {
             $values = array_diff_key($values, array_flip($options['excludeProperties']));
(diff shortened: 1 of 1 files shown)


zerocrates:
This won't work properly if `properties` contains terms that the resource has no values for: you'll get an entry for the "missing" terms in the final result when you shouldn't (despite its name, `array_replace` will introduce new keys in the result that aren't in the first arg). 

It's possible to fix this with more array_ calls... one option that comes to mind would be doing another array_intersect_keys with the original `$values` passed second to drop out anything in we don't actually have a value for... but we may just want to swap this out with a foreach:

```
$result = [];
foreach ($options['properties'] as $term) {
    if (isset($values[$term])) {
        $result[$term] = $values[$term];
    }
}
$values = $result;
```

jimsafley:
I feel that `propertiesRespectOrder` is an awkward option name. It implies that the default behavior is somehow not respecting an order. Maybe a better option is `originalOrder` with the default set to `true`? That way the default behavior remains the same, and `originalOrder=false` respects the order passed in `properties`.

zerocrates:
I'd want `properties` first in any option as this is something that only works with the `properties` option. `propertiesOrdered` or whatever if we want to shorten it, i'm fine with. The reverses, like `propertiesOriginalOrder` also fine, but also kind of long.

application/src/Api/Representation/AbstractResourceEntityRepresentation.php
```diff
@@ -481,14 +481,19 @@ public function displayValues(array $options = [])
...
+        $options['propertiesRespectOrder'] ??= false;
         $options['excludeProperties'] ??= [];
 
         $services = $this->getServiceLocator();
         $values = $this->values();
 
         // Filter values by the "properties" and "excludeProperties" options.
         if ($options['properties']) {
-            $values = array_intersect_key($values, array_flip($options['properties']));
+            $termsAsKeys = array_flip($options['properties']);
```
jlayt, line 492: One issue with doing this here is that some Modules that process 'rep.resource.display_values' don't respect the display order passed in, they quite often just re-sort it to the default template order. The sorting is probably best done after the event returns the values.

zerocrates: Do you have specific ones in mind?

jlayt: Yeap, Advanced Resource Templates does it. I've opened an issue for it over there to respect the order passed through, but it does feel like it would be whack-a-mole.

PR #2419, opened 2026-01-26
add php8.5 and fix deprecation

tacman:


zerocrates:
So, I think we'd like this instead as just `$format &&` stuck onto the front of existing `if` check, no extra `if`.

And if you could squash/rebase this down to two commits, one adding to the test matrix and one making your fix, that would be great.

Issue #2428, opened 2026-01-30
Property creation via API failing with 422 error


tacman:
## Summary
Creating properties via the Omeka S REST API fails with `422` and the error `o:vocabulary: "A vocabulary must be set."` even when a valid vocabulary exists and is referenced in the payload. Item creation and other API endpoints work.

## Steps to Reproduce
1. Ensure a custom vocabulary exists (e.g., prefix `kpa`, id `5`).
2. Call `POST /api/properties` with payload:

```json
{
  "o:term": "kpa:school",
  "o:label": "School",
  "o:local_name": "school",
  "o:vocabulary": { "o:id": 5 }
}
```

3. Observe API response.

## Observed Behavior
- API responds with `422`
- Response body (errors):
```json
{
  "o:vocabulary": ["A vocabulary must be set."]
}
```

Tried the following variants for `o:vocabulary`, all rejected with the same error:
- integer `5`
- object `{ "o:id": 5, "o:prefix": "kpa" }`
- object `{ "@id": "https://example.com/api/vocabularies/5" }`
- full vocabulary resource object returned from `GET /api/vocabularies/5`

## Expected Behavior
Property should be created when `o:vocabulary` references an existing vocabulary.

## Environment
- Omeka S REST API
- API key can create items and templates successfully
- Example API URL: `https://kpa-omeka.survos.com/api`
- Custom vocabulary exists and is visible in the admin UI

## Question
Is property creation via API supported/allowed, and if so, what exact payload is required for `o:vocabulary`?

I'm running the release from the docker repo.  I can share my API key and access if that helps.  email me at tacman@gmail.com

PR #2429, opened 2026-01-30
Allow to set vocabulary when hydrating class and property

jimsafley:
See #2428

This change makes it possible to assign a vocabulary to a resource class or property via the API.

### DIFF
--- a/application/src/Api/Adapter/PropertyAdapter.php
+++ b/application/src/Api/Adapter/PropertyAdapter.php
@@ -65,6 +65,8 @@ class PropertyAdapter extends AbstractEntityAdapter
         ErrorStore $errorStore
     ) {
         $data = $request->getContent();
+        $isCreate = Request::CREATE === $request->getOperation();
+
         $this->hydrateOwner($request, $entity);
 
         if ($this->shouldHydrate($request, 'o:local_name')) {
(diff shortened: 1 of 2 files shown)


zerocrates:
Hmm... enabling this would make it easy to, say, update a member by moving it from one vocab to another. I'm not sure whether we want to improve the api stuff for vocabs and members individually, or instead push the RDF importer via the API and have that be the intended entry point like it is for the UI.

I haven't looked closely at this in a while but if I remember correctly these will work through the API, just only if edited/created via their parent vocab.

jimsafley:
> I haven't looked closely at this in a while but if I remember correctly these will work through the API, just only if edited/created via their parent vocab.

Running the following will delete all existing properties, replacing them with the "Foo" property. 

`PATCH /api/vocabularies/1`

```
{
    "o:property": [
        {
            "o:local_name": "foo"
            "o:label": "Foo",
        }
    ]
}
```

For a `PATCH` that's not expected and certainly destructive. The only way to safely create a property via the vocabulary endpoint is to include ID'd references to all existing properties as well. Note that when `GET`ing a vocabulary, the response does not include its members.

zerocrates:
Well it's how patch works everywhere we have it. Patch an item with one file and you'll delete the others, etc. Though yes, that the output doesn't include these may make it a more likely mistake.

The design of what we have is based on the assumption that you'll create the vocab all at once with its members. In that case it's not really an issue, but editing/adding after the fact, it doesn't offer much for that.

If creation of new members is the main concern we could hydrate the vocab only on create, that's an option.

PR #2432, opened 2026-02-09
Composer addons with override

Daniel-KM:
Hi,

This is an history rebuild for #2412 , so it adds only the main feature and then additionnal but useful features in order to manage omeka modules and themes via the composer of omeka in directory composer-addons/, and to let people to continue to use the directories modules/ and themes/ as they used to. It is very useful to manage containers with volumes and to let freedom to users. It is fully compatible with 4.2 (except the change of the directory by the internal installer plugin, that was never used except by the default theme). So you can merge all or cherry-pick the features you want. 

- manage directory composer-addons/ with composer
- manage priority for modules/ and themes/
- use composer.json when available for simple and fast init
- manage automatic inflection of installer name to simplify upgrade
- a script to install modules dependencies for manually installed modules via git inside modules/, since composer does not inherit them (can be a composer command).

So the remaining point to think about is to use, or not, composer-merge-plugin (so two separate composer.json, a little slower).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@
 /files/
 /modules/
 /themes/
+/composer-addons/modules/*
+!/composer-addons/modules/.gitkeep
+/composer-addons/themes/*
+!/composer-addons/themes/.gitkeep
 /application/test/config/database.ini
 /application/test/.phpunit.result.cache
 /application/asset/css/font-awesome/
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -50,6 +50,7 @@ return $config
             ->exclude('application/data/doctrine-proxies')
             ->exclude('application/data/media-types')
             ->exclude('application/data/overrides')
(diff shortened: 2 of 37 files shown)


C-Norton:
We're interested in Bot-Challenge, so we're interested in seeing this go into a mainline omeka release at some point, as from my understanding it is necessary

Daniel-KM:
No, you can use Bot Challenge independantly of this pr. Unlike Atom where the bot challenge is integrated in the core, Bot Challenge is a module, but if this pr is merged, it will be easy to install Bot Challenge in a default project.

C-Norton:
> No, you can use Bot Challenge independantly of this pr. Unlike Atom where the bot challenge is integrated in the core, Bot Challenge is a module, but if this pr is merged, it will be easy to install Bot Challenge in a default project.

Ah, thanks for the clarification, glad to hear it

Daniel-KM:
Rebased on v4.2.1.

PR #2439, opened 2026-02-19
Fixed check of matched route in cli.

Daniel-KM:
There were some issues on the forum on [CleanUrl that break csv import](https://forum.omeka.org/t/csv-import-error-call-to-a-member-function-getparam/28675/4). So this patch avoids to prepare the route in cli context.

### DIFF
--- a/application/src/Mvc/Status.php
+++ b/application/src/Mvc/Status.php
@@ -74,8 +74,9 @@ class Status
     public function getRouteMatch()
     {
         // Attempt to get the route match from the MVC event.
+        // There is no route in cli (background jobs) because no http request.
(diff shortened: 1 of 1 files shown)


Daniel-KM:
But, unlike the previous pr like [composer](https://github.com/omeka/omeka-s/pull/2432), this fix may have complex implications with modules that rely on matched routes during jobs, so it should be verified in various contexts. Anyway, i included a similar check in module CleanUrl too.

Daniel-KM:
This is not only CleanUrl that breaks background job : there is Metadata Browse too, that does not check it. The fix in https://github.com/omeka-s-modules/MetadataBrowse/pull/54 is not yet released in a new version.

PR #2443, opened 2026-02-23
Used command "magick" over deprecated or removed "convert".

Daniel-KM:


zerocrates:
Yeah, I don't mind shifting to `magick` especially as newer ImageMagicks will now complain about using `convert`.

I'm just going to have to try to do a little research on how far back we can expect `magick` to be there and whether we should be using `magick convert` vs. `magick`... my guess is that we're not doing anything complex enough to where the differences in syntax will matter. But I'll double-check. Users of GraphicsMagick might also be another problem case, but I think we already don't work with GraphicsMagick, and we don't purport to.

Of course the way you're doing it here will fall back to `convert` so it won't be particularly problematic (assuming there are no issues with the 6 vs. 7 syntax).

Daniel-KM:
yes, there is no emergency, main distributions will keep a link to convert in next releases, so some decades.

PR #2452, opened 2026-03-11
Add experimental SQLite support to enable in-browser WASM Omeka S playground

erseco:
This PR adds **experimental SQLite support** to Omeka S.

This work was not done as a convenience change. It became necessary to build an **Omeka S Playground** comparable in spirit to **WordPress Playground**: a version of Omeka S that runs **entirely in the browser through WebAssembly**, with no traditional server-side installation required.

That playground is already being used here:

- live Omeka S WASM demo: https://ateeducacion.github.io/omeka-s-playground 
- [source code](https://github.com/ateeducacion/omeka-s-playground)

Why this matters:

- It makes Omeka S much easier to try, teach, demo, and test.
- It lowers the barrier for first contact with the platform.
- It enables ephemeral and reproducible environments for experimentation.
- It opens the door to browser-based demos, workshops, documentation examples, and QA scenarios.

This SQLite support is the foundation that makes that possible in a browser/WASM runtime, where MySQL is not a practical fit.

A related benefit is support for `blueprint.json`-style setup files in the playground, which makes it possible to preload modules, themes, users, and configuration in a simple and reproducible way.

This PR is intentionally framed as **experimental**. The goal is not to replace MySQL/MariaDB for standard production deployments, but to make new execution environments possible for testing, onboarding, demos, and browser-based experimentation.


Daniel-KM:
Yes, very interesting. Omeka works fine with PostgreSql too, so something more generic to use any sql engine would be very useful.
Furthermore, i was thinking about a way to share config between installs, so why not blueprint.json. Anything that make omeka simpler to manage is welcomed. I published a pr #2432 to use composer to install omeka, while keeping process compatible with the current by-module way.

alexdryden:
@erseco this looks really cool! I was wondering if this could be useful for storing/playing archived sites? For example, our dream would be a world where we could archive a site in a repository in such a way that users could download a set of files and see the site in a form that was as close to its original publication as possible. 

Am I right to think that this pr would get us a long way to that kind of solution?

erseco:
Yes @alexdryden, the WASM runtime environment would make something like that relatively straightforward.

That said, this PR is about one of the foundational pieces for that: SQLite support. To avoid mixing broader product/design discussions into this PR, I’d suggest continuing the conversation either in the [Omeka forum](https://forum.omeka.org/t/run-omeka-s-in-your-browser-with-wasm/29027/7) or as an issue in the [omeka-s-playground repository](https://github.com/ateeducacion/omeka-s-playground).

Cheers!

PR #2477, opened 2026-04-21
Replace Laminas\Math and Laminas\Dom with native PHP

jimsafley:
Replaces `Laminas\Math\Rand` and `Laminas\Dom\Query` with native PHP equivalents, reducing dependency on third-party packages.

- `Laminas\Math\Rand::getBytes()` and `::getString()` replaced with `random_bytes()` and `random_int()`, which provide identical cryptographic security guarantees
- `Laminas\Dom\Query` replaced with `DOMDocument` + `DOMXPath`, preserving identical parsing and query behavior

After merging, `laminas/laminas-math` and `laminas/laminas-dom` should be removed from `composer.json` if no longer pulled in transitively.

### DIFF
--- a/application/src/Entity/ApiKey.php
+++ b/application/src/Entity/ApiKey.php
@@ -2,7 +2,6 @@
 namespace Omeka\Entity;
 
 use DateTime;
-use Laminas\Math\Rand;
 
 /**
  * @Entity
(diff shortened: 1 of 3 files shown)


zerocrates:
OK just one more really minor thing, let's `use` `DOMDocument` and `DOMXPath` rather than using their FQCNs directly.

PR #2478, opened 2026-04-22
Validate vocabulary prefix against allowed character pattern

jimsafley:
The vocabulary prefix is user-supplied but was never validated against the character pattern that `isTerm()` already enforces. This meant a vocabulary could be created with an invalid prefix that would cause term lookups to silently fail.

This change adds prefix format validation at both the form level (for immediate user feedback) and the API adapter level (as a server-side guard for direct API writes). The allowed character pattern is centralized so it isn't duplicated across files.

A pre-existing bug in `isTerm()`'s regex is also fixed: the hyphen in `[a-z0-9-_]` was in an ambiguous position that caused it to match a broader range of characters than intended. It is moved to the end of the character class: `[a-z0-9_-]`.

### DIFF
--- a/application/src/Api/Adapter/AbstractEntityAdapter.php
+++ b/application/src/Api/Adapter/AbstractEntityAdapter.php
@@ -10,6 +10,7 @@ use Omeka\Api\Response;
 use Omeka\Db\QueryBuilder as OmekaQueryBuilder;
 use Omeka\Entity\User;
 use Omeka\Entity\EntityInterface;
+use Omeka\Entity\Vocabulary;
 use Omeka\Stdlib\ErrorStore;
 use Laminas\EventManager\Event;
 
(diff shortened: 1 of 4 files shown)


jimsafley:
This should fix #2478

zerocrates:
Though it's tempting to reuse the pattern in pieces like this, because the prefix is part of the term pattern, here I think it's best not to do that. My concern with splitting the pattern up is that it separates it out from the flags and other parts of the pattern, which to me makes it harder to read what the actual pattern is and makes it more likely that some later usage of these partial pattern constants accidentally messes up the flags or start/end assertion characters.

My preference would be to have the prefix pattern regex constant be the whole pattern string to be used as-is. For this feature we'd only really need that one, but having also a second constant that's the entire "term" pattern (just the exact string we were using in isTerm) is a fine idea, too. Having the two full patterns as constants defined right next to each other should provide enough future clarity that the two patterns are related.

PR #2482, opened 2026-04-24
Add sortable column headers to admin browse pages

### LINKED ISSUE #2420: Indicate sort order in tables with sortable columns
In tables with sort functionality, the currently sorted column should include an `aria-sort` attribute to indicate sort order for assistive tech users. Is this possible in the current pattern since sorting does not happen dynamically, or would that have no bearing?

### LINKED ISSUE #2456: Improvement: Table headers should be clickable to change sort field and order
On browse pages (items, item sets, media), the table column headers are not interactive. There is no way to click on a header to sort by that column, and the headers do not visually indicate which column is currently being sorted or in which direction.

Clicking a table column header should set that column as the active sort field. Clicking the same header again should toggle between ascending and descending order. The currently active sort field (if part of the visible table columns) should be visually indicated, for example with an arrow icon showing the sort direction. 

Clickable, sortable table headers are a well-established convention that users have come to expect from virtually any data table interface. This would make browsing and organizing resources much more intuitive.

jimsafley:
Previously, sorting on browse pages required using the sort dropdown above the table. Column headers were plain text with no interactive behavior.

Headers are now clickable sort links. Clicking a header sorts the list by that column; clicking again toggles between ascending and descending order. An arrow icon on the active column shows the current sort direction. This behavior is consistent across all admin browse pages: items, item sets, media, resource templates, vocabularies, jobs, assets, users, sites, and site pages.

Should fix #2456
Should fix #2420


zerocrates:
The conflict here is no big deal, we can just regenerate the CSS when merging.

Small style things:

- `a` is already defined with no text-decoration so the text-decoration:none for these links seems unnecessary
- white-space: nowrap can cause problems if the label is too long to fit in the column. Dropping it means the sort indicator can be at the end of the wrapped text but that's probably preferable to the heading just extending out of its column and you not being able to see the icon at all..

On `aria-sort`: though `none` exists as a value I think it's more typical to just not have `aria-sort` on the heading at all if its not the active sort.

This PR will conflict also with the other PR about adding column options for users; it looks like the easier path is to merge this one then that one.

jimsafley:
All three addressed:

- `text-decoration: none` dropped, redundant with the global `a` rule.
- `white-space: nowrap` dropped. Long headers wrap now, with the indicator trailing the last line.
- `aria-sort` omitted when the column isn't the active sort, and also when `sort_order` is absent, which previously announced "descending" while the query actually sorted ascending.

I also consolidated header markup into a new `Browse::renderHeaderCell()`, so `aria-sort` and the sort link are derived in one place instead of from separate calls in each template. I kept it as a second method rather than folding it into `renderHeaderRow()`: only four browse tables draw their columns from config, and the other eight have no resource type to pass, so a single method would mean a placeholder argument doing nothing at two thirds of the call sites. `renderHeaderRow()` keeps its signature and delegates to the new method.

jimsafley:
One note on the new method: omit the `$sortBy` argument and you get a non-sortable header. That's how the asset browse renders its "Owner" column.

PR #2496, opened 2026-05-20
Add item count browse column for item sets

### LINKED ISSUE #2493: Add item set browse column: Item Count
Image

In the user settings, there is no option to add a browse column for item set item count. This is highly useful information that should be relatively easy to add

jimsafley:
Adds an "Item count" browse column for item sets, configurable in user settings. The column displays the number of items in each set as a link to the filtered items browse. Sortable by item count.

Note that adding this column will result in one additional COUNT query per item set on the browse page. Since this is an opt-in column, users can simply remove it if performance is a concern.

Fixes #2493


C-Norton:
Thanks Jim!

PR #2505, opened 2026-06-04
Add mbstring as requirement

Sjord:
And update links.

During installation Omeka S will complain:

Omeka requires the PHP extension mbstring, but it is not loaded.

So better to make this clear in the README.


zerocrates:
We've added mbstring to the list in the manual, and I'm going to just point the readme there so this information is only in one place.

PR #2511, opened 2026-06-08
Move Welcome page creation to event listener, add to navigation

jimsafley:
New sites include a Welcome page but it was never added to navigation, requiring manual assignment. This moves Welcome page creation out of `SiteAdapter::hydrate()` and into an `api.create.post` event listener, where the page ID is available after flush. The listener creates the page and adds it to navigation in one step.

### DIFF
--- a/application/Module.php
+++ b/application/Module.php
@@ -115,6 +115,12 @@ class Module extends AbstractModule
         );
 
         $sharedEventManager->attach(
+            'Omeka\Api\Adapter\SiteAdapter',
+            'api.create.post',
+            [$this, 'addSiteWelcomePage']
+        );
+
+        $sharedEventManager->attach(
             '*',
             'api.context',
             [$this, 'addTermDefinitionsToContext']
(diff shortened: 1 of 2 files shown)


kimisgold:
I'd like to get some input here from @aviannamiller, @allanaaa, and @katknow. This pull request automatically adds the "Welcome" page to a new site's navigation; previously, new sites would default to the item browse and the Welcome page was inaccessible unless manually added to the navigation. In this change, new sites would default to a "This site has no pages" view for the landing page.

Is this an improved workflow? My thinking is that the Welcome page is pure boilerplate, and that a user would want to start a landing page from scratch. I've heard arguments that it's common for site admin to start building a landing page by editing the Welcome page. I could definitely see that, but in any case, it would be good to have some other perspectives on the site creation process.

aviannamiller:
I don't feel too strongly either way, but I will say I do not particularly mind having the landing page default to the Welcome page. It feels pretty familiar in terms of expectations when starting out with a site builder. Easy enough to wipe out or change the landing page, as well, I suppose.

If going with this route, I am wondering if it is possible to have the Welcome page loaded in as the selection in the dropdown to reflect what is set. Currently, it displays with "First page in navigation":
'Select a homepage' input, selection as described above Expanded 'Select a homepage' input, selection as described above, with secondary choice of 'Welcome' page available in dropdown

allanaaa:
I don't understand. The new workflow is - the Welcome page is added to the navigation, but is not selected as the homepage - the homepage is also not Browse items - it would display a "This site has no pages" despite there being one page? Or do you mean that that last part is the old workflow and the new one would not ever display that?

I think we should have the default homepage be Browse Items on all new sites. "This site has no pages" is not useful, especially if there's no indication on the admin side that that's going to happen. 

I slightly prefer to **not** add the Welcome page to the navigation. It's meant to be a page with example content, it doesn't need to be seen by the public. Someone editing the page can View it in public without it being in the navigation.

zerocrates:
OK just to make sure everyone is on the same page...

What happens currently is:
- The Welcome page is created, and is publicly visible, but isn't in the navigation
- The Browse page is in the navigation
- The default homepage is "This site has no pages" because only "real"/custom Pages are pickable as the homepage

The homepage algorithm picks the first "real" page in the nav if there isn't one specifically selected. In either case only Pages can be the homepage; Browse isn't and can't be the home.

This PR puts Welcome in the nav, which means it would be the automatic homepage without having to set it as the explicit homepage.

Changing how homepages work to allow more page types is a possibility; it would be a whole different PR.

allanaaa:
If we were going to automatically make a Welcome page accessible it shouldn't be this demonstration of styles for the admin; it should be a bland welcome message (and added to translateable strings, probably). 

I still think a boilerplate page with styles is a good idea, but not appropriate content for this use. 

I'm fine with either: 
- adding a second built-in page that will be the Welcome page (published and set as the homepage in the nav) and leaving this boilerplate page as-is (or setting it to private), or 
- making Browse Items the default.

PR #2519, opened 2026-06-17
Allow to use http/2 when available

Daniel-KM:


zerocrates:
We're already pinned to laminas-http 2.22.0 currently so I don't think you actually need any composer changes.

On defaulting to curl if the support is there, I don't think I mind, and triggering automatic choosing when the configured adapter is null seems like a good way to go about it.

I'm not sure I want it to auto-drop to Socket if you explicitly choose curl and don't have support, though. Just letting it error might be a better move in that situation.

Daniel-KM:
Ok, explicit curl or socket or any other adapter does not fallback to socket anymore.

Note that the version of laminas 2.22 is a fix i made in 2025 (https://github.com/laminas/laminas-http/pull/109) and i have some other improvements in omeka and modules that i use since some years but that i didn't even push any more.

zerocrates:
On laminas-http my point was just that Omeka S already comes with 2.22.0; it's been the version in use since version 4.2.0 from last December.

PR #2522, opened 2026-06-22
Assign new item set

Daniel-KM:
This is something that my users ask regularly, like for items.

So i reimplemented the third solution of the pr #2057 , but finally i think that the best solution will be another one, so to replace o:assign_new_items by a simple setting and simplify the table site, but it is another thing.

### DIFF
--- a/application/src/Api/Adapter/ItemSetAdapter.php
+++ b/application/src/Api/Adapter/ItemSetAdapter.php
@@ -1,6 +1,7 @@
 <?php
 namespace Omeka\Api\Adapter;
 
+use Doctrine\Common\Collections\ArrayCollection;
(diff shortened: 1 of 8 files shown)


Daniel-KM:
About copilot, i think the setting is on your part, not mine.

sharonmleon:
Our settings are all off. The only way this works is if someone is paying for it and we definitely are not.

Daniel-KM:
Of course, i don't have any subscription at github, i don't have windows, i don't have any microsoft product.
Or github may have categorized me as open-source writer, and maybe it is automatically set copilot for free in that case. 
I'm going to search more, but the options are very numerous and ununderstandable. There is no a unique option "Disable copilot" (the same for "Respect gdpr" anyway).

PR #2529, opened 2026-07-10
Add a secret key in the config.

Daniel-KM:
All cms have one or multiple random encrypt keys, for example to protect a api key (in particular modules AiGenerator and Translator or a module that will follow more strictly the grpd). 
So this pr adds it in omeka (config/secret_key.php and env) and a form element to use it.

I disabled copilot on pr, but i'm not sure i can because it seems to be enabled even if it is disabled, so it is a check for that too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@
 .DS_Store
 /application/language/debug.*
 /static/
+/config/secret_key.php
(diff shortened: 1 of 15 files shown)


Daniel-KM:
So the copilot is disabled !

image

Daniel-KM:
Note that  the form element works only in ConfigForm for now, it needs some specific code to be used in settings.

Issue #2396, opened 2025-12-17
ThumbnaiRenderer does not set tag download attribute


jlayt:
The ThumbnailRenderer does not set the download tag on the file link, so the file downloads using the internal file reference and not the real source file name.

https://github.com/omeka/omeka-s/blob/4bb3350dad55fc50ce7b4370bec978ba979b968e/application/src/Media/FileRenderer/ThumbnailRenderer.php#L25-L27

This is easily amended to add the download attribute so the file gets correctly named.

```
    $title = $media->displayTitle();
    $source = $media->source();
    return sprintf('<a href="%s" download="%s" title="%s">%s</a>', $view->escapeHtml($url), $view->escapeHtml($source), $view->escapeHtml($title), $img);
```

PR #2537, opened 2026-07-27
Set Download parameter for ThumbnailRenderer file link

jlayt:
This resolves issue  #2396 by setting the download tag on the ThumbnailRenderer file link, so the file download uses the original file name and not the internal media resource ID.


zerocrates:
The problem with doing this is that it sets all those links to download, so you'll download an original image rather than view it in the browser. I'm not sure we want to do that....

PR #2553, opened 2026-08-21
changed how $owner is found for Job/Dispatcher.php

GarrettArm:
This change is intended to resolve this error when running an AdvancedSearch\Job\IndexSearch.

In the GUI, I navigate to /admin/search-manager then click "ReIndex" on my Search Engine: "solr".  Here are my module versions pinned in a Dockerfile:  https://github.com/uncw-library/omeka-s.git

The error:

2026-08-21T15:55:21+00:00 ERR (3): Doctrine\ORM\ORMInvalidArgumentException: A new entity was found through the relationship 'Omeka\Entity\Job#owner' that was not configured to cascade persist operations for entity: DoctrineProxies_CG_\Omeka\Entity\User@1071. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example #[ORM\ManyToOne(..., cascade: ['persist'])]. If you cannot find out which entity causes the problem implement 'Omeka\Entity\User#__toString()' to get a clue. in /var/www/html/vendor/doctrine/orm/src/ORMInvalidArgumentException.php:103

This pull request resolves that error.
Please let me know of any thoughts or questions,
Garrett Armstrong

Full error message:
2026-08-21T17:56:49+00:00 DEBUG (7): Process done with the main entity manager {"referenceId":"search/index/job_969"}
2026-08-21T17:56:49+00:00 NOTICE (5): Search index #2 ("solr"): start of indexing {"referenceId":"search/index/job_969"}
2026-08-21T17:56:53+00:00 INFO (6): Search index #2 ("solr"): end of indexing. item_sets: 9 indexed; items: 266 indexed. Execution time: 4 seconds. Failed indexed resources should be checked manually. {"referenceId":"search/index/job_969"}
2026-08-21T17:56:53+00:00 INFO (6): End of indexing. Execution time: 4 seconds. Max memory: 46137344. Failed indexed resources should be checked manually. {"referenceId":"search/index/job_969"}
2026-08-21T17:56:53+00:00 ERR (3): Doctrine\ORM\ORMInvalidArgumentException: A new entity was found through the relationship 'Omeka\Entity\Job#owner' that was not configured to cascade persist operations for entity: DoctrineProxies\__CG__\Omeka\Entity\User@1071. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example #[ORM\ManyToOne(..., cascade: ['persist'])]. If you cannot find out which entity causes the problem implement 'Omeka\Entity\User#__toString()' to get a clue. in /var/www/html/vendor/doctrine/orm/src/ORMInvalidArgumentException.php:103
Stack trace:
#0 /var/www/html/vendor/doctrine/orm/src/UnitOfWork.php(3864): Doctrine\ORM\ORMInvalidArgumentException::newEntitiesFoundThroughRelationships()
#1 /var/www/html/vendor/doctrine/orm/src/UnitOfWork.php(418): Doctrine\ORM\UnitOfWork->assertThatThereAreNoUnintentionallyNonPersistedAssociations()
#2 /var/www/html/vendor/doctrine/orm/src/EntityManager.php(414): Doctrine\ORM\UnitOfWork->commit()
#3 /var/www/html/application/src/Job/Dispatcher.php(85): Doctrine\ORM\EntityManager->flush()
#4 /var/www/html/modules/AdvancedSearch/src/Job/IndexSearch.php(776): Omeka\Job\Dispatcher->dispatch()
#5 /var/www/html/modules/AdvancedSearch/src/Job/IndexSearch.php(303): AdvancedSearch\Job\IndexSearch->reindexSuggesters()
#6 /var/www/html/application/src/Job/DispatchStrategy/Synchronous.php(34): AdvancedSearch\Job\IndexSearch->perform()
#7 /var/www/html/modules/Common/src/Job/Dispatcher.php(27): Omeka\Job\DispatchStrategy\Synchronous->send()
#8 /var/www/html/application/data/scripts/perform-job.php(66): Common\Job\Dispatcher->send()
#9 {main} {"referenceId":"search/index/job_969"}

### DIFF
--- a/application/src/Job/Dispatcher.php
+++ b/application/src/Job/Dispatcher.php
@@ -80,7 +80,11 @@ class Dispatcher
         $job->setStatus(Job::STATUS_STARTING);
         $job->setClass($class);
         $job->setArgs($args);
-        $owner = $this->auth->getIdentity(); if ($owner) { $owner = $this->entityManager->getReference(\Omeka\Entity\User::class, $owner->getId()); } $job->setOwner($owner);
+        $owner = $this->auth->getIdentity();
+        if ($owner) {
+            $owner = $this->entityManager->getReference(\Omeka\Entity\User::class, $owner->getId());
+        }
+        $job->setOwner($owner);
         $this->entityManager->persist($job);
         $this->entityManager->flush();
 
--- a/application/src/Job/Dispatcher.php
+++ b/application/src/Job/Dispatcher.php
@@ -80,7 +80,7 @@ class Dispatcher
         $job->setStatus(Job::STATUS_STARTING);
         $job->setClass($class);
         $job->setArgs($args);
-        $job->setOwner($this->auth->getIdentity());
+        $owner = $this->auth->getIdentity(); if ($owner) { $owner = $this->entityManager->getReference(\Omeka\Entity\User::class, $owner->getId()); } $job->setOwner($owner);
         $this->entityManager->persist($job);
         $this->entityManager->flush();
 


zerocrates:
The problem here happens because that job calls `clear`, which can cause this kind of problem with relationships like this one between the job and its owner/starter. The Advanced Search module's job reloads the Job record after clearing, but not the user. We often try to avoid this kind of problem by not calling `clear` but instead detaching "new" entities specifically, which is a change that could happen on the module side and also fix this.

The other bug you mention, it's a similar but not quite the same issue.

On what your change actually does: I don't think we've explored using a reference/proxy on purpose to avoid these issues in a different way like you're doing here but it's an interesting possibility. The Doctrine checker that raises this error just ignores proxies entirely, which is convenient.

For jobs I think we might actually go further and use the database connection directly, taking the job status and logging operations out of the ORM entirely, but that's a bigger change and I don't think it necessarily stands in the way of something smaller like this.

On code style: we don't do multiple statements in a line like this; it should be split out into all separate lines.

zerocrates:
Did you change any other settings or anything when you were testing this change out?

Since you're changing the `dispatch` method, I don't actually see how it could fix anything or make a difference generally in a background job, since `dispatch` doesn't get called when a background job is running.

GarrettArm:
I appreciate you sharing that context.  Honestly, I'm not sure why this change works and php is not my strongest language.  An LLM suggested I try this approach.  There was no change to the setup except this one line, and the outcome was repeatable.  

With the original code in place, the reindexing job would give the error.  Then after replacing it with this version, the job would complete.  And back again.

My code revision may even be bizarre, since I'm not really sure why it would work.  But it seems to work.  If the approach is going in the completely wrong direction, feed free to close this pull request.  Maybe it points toward a solution though??

Garrett

